PR #4, opened 2023-11-13
Rust base refactor

algebraic-sofia:
- Cargo fmt and Cargo clippy --fix
- Changed u64 to u128 in the Merkle tree in order to make it easier to build big commitment hashes


offchain/core/src/machine/rpc.rs
```diff
@@ -181,7 +163,7 @@ fn add_and_clamp(x: u64, y: u64) -> u64 {
 
 fn encode_access_log(log: &AccessLog) -> Vec<u8> {
     let mut encoded = Vec::new();
-    
+
     for a in log.accesses.iter() {
         assert_eq!(a.log2_size, 3);
         if a.r#type == AccessType::Read {
```
GCdePaula, line 169: I've no idea what r#type is 😱

offchain/core/src/machine/rpc.rs
```diff
@@ -181,7 +163,7 @@ fn add_and_clamp(x: u64, y: u64) -> u64 {
 
 fn encode_access_log(log: &AccessLog) -> Vec<u8> {
```
GCdePaula, line 164: Can you take a closer look in this function? We can talk to explain what's going on here

Take a look at the jsonRPC interface [here](https://github.com/cartesi/machine-emulator/blob/main/src/jsonrpc-discover.json). You can paste that file [here](https://playground.open-rpc.org) to get a nice view. Go to `step_uarch` and check the access log schema. I guess it's already mapped into Rust.

This function essentially creating this big array with all the proofs. There's some annoying endianness consideration probably, and the fact that the blockchain merkle verification consumes the hashes "backwards". I'd argue the machine generates it backwards, but anyway. We reverse it at some point.

Please verify if we're doing it nicely. I believe it's correct, just wondering if the style is ok too.

offchain/core/src/merkle/digest/keccak.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Computes the Keccak256 hash of the given data and returns a new Digest.
+    pub fn hash(data: &[u8]) -> Digest {
+        let mut keccak = Keccak256::new();
+        keccak.update(data);
+        let digest: [u8; 32] = keccak.finalize().into();
+        Digest::from(digest)
+    }
+
+    /// Joins the current Digest with another Digest to create a new Digest.
+    pub fn join(&self, digest_2: Digest) -> Digest {
```
GCdePaula, line 17: Maybe `digest_2` should also be a reference?

offchain/core/src/merkle/digest/keccak.rs
```diff
@@ -0,0 +1,24 @@
+//! Keccak256 hash for the Digest Type. It's used to hash the data in the Digest.
+
+use sha3::{Keccak256, Digest as Keccak256Digest};
+
+use super::Digest;
+
+impl Digest {
+    /// Computes the Keccak256 hash of the given data and returns a new Digest.
+    pub fn hash(data: &[u8]) -> Digest {
```
GCdePaula, line 9: What do you think of the name "from_data"?

I think `Digest::from_data` gives a better understanding of `Digest::hash`.

offchain/core/src/merkle/digest/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl Digest {
+    /// Creates a new [Digest] with the provided 32-byte data.
+    pub fn new(data: [u8; 32]) -> Self {
+        Digest { data }
+    }
+
+    /// Attempts to create a [Digest] from a Vec<u8> containing 32 bytes of data.
+    pub fn from_data(digest_data: &[u8]) -> Result<Digest, Box<dyn Error>> {
```
GCdePaula, line 24: On errors, let's try using a library.

Currently, we're using both `anyhow` and `this_error`. @renan061 are those the current recommendations for error libraries in Rust?

 I think for this case (library), `this_error` is better suited.

GCdePaula: Also, the name here should be something closer to `from_digest`, or `from_hash`.

offchain/core/src/merkle/digest/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+            return Err("Invalid digest data length".into());
+        }
+
+        let mut data = [0u8; 32];
+        data.copy_from_slice(digest_data);
+        Ok(Digest::new(data))
+    }
+
+    /// Attempts to create a [Digest] from a hexadecimal string.
+    pub fn from_hex(digest_hex: &str) -> Result<Digest, Box<dyn Error>> {
```
GCdePaula, line 35: Likewise for the error library.

On names, maybe something like `from_hash_hex` or `from_digest_hex`? Something that explains that the input is already a hash.

offchain/core/src/merkle/tree_builder.rs
```diff
@@ -1,165 +1,187 @@
...
 use std::collections::HashMap;
 
-use sha3::{Digest, Keccak256};
+use crate::merkle::{Digest, MerkleTree, MerkleTreeLeaf, MerkleTreeNode};
 
-use crate::{
-    merkle::{ Hash, MerkleTreeNode, MerkleTreeLeaf, MerkleTree},
-    utils::arithmetic,
-};
+pub type Int = u128;
```
GCdePaula, line 7: Call it at least "UInt". Also, I think we should use [`ruint`](https://docs.rs/ruint/latest/ruint/index.html).

GCdePaula: I think we should also try to make the interface take a normal unsigned integer. Like, maybe something like `impl Into<UInt>` could work?

offchain/core/src/merkle/node.rs
```diff
@@ -1,33 +1,32 @@
...
+            digest,
+            children: None,
         }
     }
 
-    pub fn set_children(&mut self, left: Hash, right: Hash) {
-        self.left = Some(left);
-        self.right = Some(right);
+    /// Sets the children of the node.
+    pub fn set_children(&mut self, left: Digest, right: Digest) {
```
GCdePaula, line 24: This shouldn't be a public method.

offchain/core/src/merkle/node.rs
```diff
@@ -1,33 +1,32 @@
...
-    left: Option<Hash>,
-    right: Option<Hash>,
+    pub digest: Digest,
+    children: Option<(Digest, Digest)>,
 }
 
 impl MerkleTreeNode {
-    pub fn new(digest: Hash) -> Self {
+    /// Creates a new Merkle tree node with the given digest.
+    pub fn new(digest: Digest) -> Self {
```
GCdePaula, line 16: I think the constructor should take the children too.

GCdePaula: Or maybe two constructors: one for the inner node (takes two children and internally we join them and figure out the digest), and another for the leaf (takes just the digest).

GCdePaula (commented):
Great first PR ✨✨

I'm still reviewing it. When I get back, let's schedule a talk! I'm trying to understand some parts of it. Also, we usually name feature branches just `feature/` instead of `features/`.

Anyway, nice work! Cheers.

offchain/core/src/machine/commitment.rs
```diff
@@ -54,32 +37,32 @@ pub async fn build_big_machine_commitment(
     log2_stride_count: u64,
 ) -> Result<MachineCommitment, Box<dyn Error>> {
     let machine_lock = machine.clone();
-    let mut machine  = machine_lock.lock().await; 
-    
+    let mut machine = machine_lock.lock().await;
```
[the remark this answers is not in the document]
stephenctw: The machine's methods are `async`. Because it uses `jsonrpsee` crate method for the underlying transport. C binding methods are synchronous.
https://github.com/cartesi/machine-json-rpc/blob/main/src/client/mod.rs

Is `Mutex` a requirement for `async` methods even if we are sharing not the `machine` object over different threads?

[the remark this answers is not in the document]
algebraic-sofia: I think we cant just remove all the mutexes because of all of the async stuff.

offchain/core/src/machine/commitment_builder.rs
```diff
@@ -48,39 +39,40 @@ impl MachineCommitmentBuilder for CachingMachineCommitmentBuilder {
...
-            .or_insert_with(HashMap::new)
+            .or_default()
             .insert(base_cycle, commitment.clone());
-        
+
         Ok(commitment)
     }
 }
 
 pub struct FakeMachineCommitmentBuilder {
```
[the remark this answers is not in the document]
algebraic-sofia: I think that we cannot remove because some of the functions use FakeMachineCommitmentBuilder to simulate an adversary

offchain/core/src/machine/rpc.rs
algebraic-sofia: The prefix `r#` exists just to use the keyword `type` without having to make a prefix, suffix or something like that.

offchain/core/src/merkle/node.rs
```diff
@@ -1,33 +1,39 @@
...
-    pub fn children(&self) -> Option<(Hash, Hash)> {
-        if self.left.is_some() && self.right.is_some() {
-            let left = self.left.as_ref().unwrap();
-            let right = self.right.as_ref().unwrap(); 
-            Some((left.clone(), right.clone()))
-        } else {
-            None
-        }
+    /// Sets the children of the node.
+    pub(crate) fn set_children(&mut self, left: Digest, right: Digest) {
```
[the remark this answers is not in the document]
algebraic-sofia: It's used inside the function `join_nodes` that gets a mutable reference of a node and then set it's children. Idk if I should change from `node.set_children(left, right)` to something like `*node = MerkleTreeNode::new(left, right)`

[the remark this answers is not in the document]
stephenctw: I saw @GCdePaula also commented on this one. Having two constructors makes a lot more sense.

stephenctw:
This PR generally looks good to me. I think we can further polish each module later.

GCdePaula (approved):
Great work :)

Issue #12, opened 2023-12-22
Rust player/arena


GCdePaula:
Rewrite the rust code to have similar structures and identical behaviors to the lua prototype.

Issue #8, opened 2023-12-22
Integration test using dave-compute and lua prototype


stephenctw:
After `dave-compute` is implemented, create a integration test that have two players fighting over a tournament. One player being the `dave-compute` node, the other one being the lua prototype node.

PR #23, opened 2023-12-25
A rust based Dave compute implementation

stephenctw:
Tackle issue: #8 and #12 
The implementation uses a similar structure to the lua prototype.


offchain/core/src/merkle/tree_builder.rs
```diff
@@ -61,67 +84,75 @@ impl MerkleBuilder {
     pub fn build(&mut self) -> MerkleTree {
         let last = self.leafs.last().expect("no leafs in merkle builder");
         let count = last.accumulated_count;
-        let mut log2_size = UInt::BITS;
+        let mut log2_size = 64;
```
algebraic-sofia, line 87: I think that if UInt is different from U64 then it will probably generate an incorrect output by assuming that the log_size is always 64.

offchain/input-reader/src/lib.rs
```diff
@@ -162,3 +157,27 @@ impl PartitionProvider {
         false
     }
 }
+
+#[tokio::test]
+
+async fn test_input_reader() -> Result<(), Box<dyn std::error::Error>> {
```
algebraic-sofia, line 163: Should this test be here? These structures to read the input-box are not being used at all in this repository.

stephenctw: The input-reader module should be used by the rollups-dave-node, which is not being implemented yet.

offchain/core/src/merkle/tree_builder.rs
stephenctw: Good point. I honestly don't know all the details of the implementation. I was referring to the lua-node when I tried to get them work for integration test.

offchain/input-reader/src/lib.rs
algebraic-sofia: Yeah, I think that the idea is that I will implement a module on top of dave that will interact with input-box and the related things, this is why I'm asking if it should or not be here instead of being in the rollup node.

algebraic-sofia: I wrongly assumed that the test was in the wrong module. It is perfectly fine to be where it is right now.

GCdePaula:
What are these large json files?

offchain/core/src/arena/arena.rs
```diff
@@ -1,23 +1,21 @@
...
+    machine::{constants, MachineProof},
     merkle::{Digest, MerkleProof},
 };
 
-/// Type alias for Ethereum addresses (20 bytes).
-pub type Address = H160;
-
 /// The [Arena] trait defines the interface for the creation and management of tournaments.
 #[async_trait]
 pub trait Arena: Send + Sync {
```
GCdePaula, line 15: Do we need a trait?

offchain/core/src/contract/factory/tournament_factory.rs
```diff
@@ -339,13 +339,13 @@ pub mod tournament_factory {
...
...\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0`\x01`\x01`\xA0\x1B\x03\x16c\\\xD8p\x16\x8B\x8B\x8B\x8B\x8B\x8B\x8B\x8B3`@Q\x8Ac\xFF\xFF\xFF\xFF\x16`\xE0\x1B\x81R`\x04\x01a\x02@\x99\x98\x97\x96\x95\x94\x93\x92\x91\x90a\x03\xCEV[`\0` \x82\x84\x03\x12\x15a\x03\x02W`\0\x80\xFD[P5\x91\x90PV[g\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\x81\x16\x81\x14a\x03\x1FW`\0\x80\xFD[PV[`\0\x80`\0\x80`\0\x80`\0\x80a\x01\0\x89\x8B\x03\x12\x15a\x03?W`\0\x80\xFD[\x885\x97P` \x89\x015\x96P`@\x89\x015\x95P``\x89\x015\x94P`\x80\x89\x015\x93P`\xA0\x89\x015a\x03m\x81a\x03\tV[\x92P`\xC0\x89\x015\x91P`\xE0\x89\x015a\x03\x84\x81a\x03\tV[\x80\x91PP\x92\x95\x98P\x92\x95\x98\x90\x93\x96PV[`\x01`\x01`\xA0\x1B\x03\x81\x16\x81\x14a\x03\x1FW`\0\x80\xFD[`\0` \x82\x84\x03\x12\x15a\x03\xBCW`\0\x80\xFD[\x81Qa\x03\xC7\x81a\x03\x95V[\x93\x92PPPV[\x98\x89R` \x89\x01\x97\x90\x97R`@\x88\x01\x95\x90\x95R``\x87\x01\x93\x90\x93R`\x80\x86\x01\x91\x90\x91Rg\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\x90\x81\x16`\xA0\x86\x01R`\xC0\x85\x01\x91\x90\x91R\x16`\xE0\x83\x01R`\x01`\x01`\xA0\x1B\x03\x16a\x01\0\x82\x01Ra\x01 \x01\x90V\xFE\xA2dipfsX\"\x12 \xB9l\xF1\t\xC3\x9B\xFC\xB7\x14\xACK\x85\xAE\xE7\x13\xCF\x0BF\xECvVL\x137\xB4]\xB7\xA5\xAA\xFA\x8C_dsolcC\0\x08\x11\x003";
```
GCdePaula, line 348: I'm not sure we should commit generated binding files.

offchain/core/src/arena/arena.rs
GCdePaula: It seems it's like an IO object thing, which we use to read from and write to the blockchain. It's possible we eventually implement Dave in non-EVM stuff, but I think it's not a concern now. But it may make it a lot easier to test things, so let's evaluate.

I see this `Arena` trait is injected into `strategy`. I think we should explore splitting this trait into two different objects: a read and a write object.

The read object it's possible we could not inject. Instead, we just read the entire state of Dave before calling react, and pass this state as an argument to react. The downside of that is that we may be reading more state than we want. It's kinda like a tree/graph traversal; we're just interested in a single path, and we need the logic in strategy to decide what we need to read. However, because of the `gc`, I think we'll need the full state anyway. So maybe this downside doesn't apply.

The write object I think we need to inject, for testing purposes at least. Like, we could return a reaction, but I'm not particularly fond of that design.

stephenctw:
> What are these large json files?

They are contract abis, I'll remove them and improve the overall binding generation process.

offchain/core/src/merkle/tree_builder.rs
stephenctw: @GCdePaula could you take a look at @algebraic-sofia's comment? I don't have enough knowledge to determine which one is correct.

GCdePaula: Let me think.

Following the logic, it would be the number of bits in our counter. But this particular piece of code was doing an overflow trick on the counter. In Lua, the number of bits was 64. This means we can represent the numbers ranging from `0` to `2^64 - 1`.

The issue is that in Lua we needed to store the number `2^64` on this counter. The solution was this: since the counter can't be zero (we need at least one node in the tree), the value `zero` represented `2ˆ64`, and all the others represented their normal value.

I think this trick was removed by just using a larger counter. So I think we can actually remove this piece of the code.

GCdePaula: So it would just be `count.trailing_zeros()`.

stephenctw: Thanks @GCdePaula , I can confirm this code works fine with integration test.
```
        assert!(count.is_power_of_two(), "is not a power of two {}", count);
        let log2_size = count.trailing_zeros();
```

GCdePaula: Yes!

offchain/core/src/arena/arena.rs
stephenctw: So a trait for the write object, no trait for the read object?

offchain/core/build.rs
GCdePaula, line 1: Forge has a the `forge bind` command, which generates a Rust crate with the bindings. I think we should create a crate just with the bindings. It's possible that it's easier to use `forge bind` for that as well. Guilherme generated Rust bindings for `rollups-contracts` using this.

offchain/core/src/strategy/player.rs
GCdePaula, line 1: I like the way you've changed the arena/player interfaces, and the way you more explicitly split the reading and sending functionalities. I'll propose one further change, let me know what you think.

My understanding is that the `Arena` object has a bunch of reading methods. These methods are called beforehand, to fetch the state. Then, we call react on `Player`, passing this state. Furthermore, `Arena` also implementes the `ArenaSender` trait, and the `Player` stores it internally.

One suggestion is to really split the `Arena` object in two. One for reading states, that doesn't really implement any trait. Something like `StateReader`, or `ArenaReader`, or some better name. Then there's the `ArenaSender` trait, which is implemented by a different object. I believe this split is already kinda there, I'm suggesting we really split them in different objects and files. I may be missing something that makes this impossible.

Also, there's another interesting design pattern in Rust which may be interesting to experiment. Currently, we have something like:

```
trait ArenaSender {}

struct Player<A: ArenaSender> {
    arena_sender: A
}

impl<A: ArenaSender> Player<A> {
    fn new(arena_sender: A) -> Self {
        Self { arena_sender }
    }

    fn react() {}
}
```

Instead, we can do:

```
trait ArenaSender {}

struct Player {}

impl Player {
    fn new() -> Self {
        Self {}
    }

    fn react(arena_sender: impl ArenaSender) {}

    // or
    fn react<A: ArenaSender>(arena_sender: A) {}
}
```

GCdePaula (commented):
This is looking super nice!!

offchain/core/src/strategy/player.rs
```diff
@@ -0,0 +1,375 @@
...
+            machine_factory,
+            machine_path,
+            commitment_builder,
+            root_tournamet,
+        }
+    }
+
+    pub async fn react(
+        &mut self,
+        tournament_states: HashMap<Address, TournamentState>,
```
GCdePaula, line 46: I think we also could create a type just for this.

offchain/core/src/strategy/player.rs
stephenctw: I'll certainly split the reader and the sender in different files and objects. I'm curious about what advantage you think we can benefit from the second design pattern? Is it that the `Player` object is no longer bundled to a specific `ArenaSender` implementation? So the `Player` can freely `react` using different `ArenaSender` implementation (possible mocking version for testing purpose) at runtime?

offchain/core/build.rs
stephenctw: Sure, `forge bind` seems much easier to do. I was maintaining this `build.rs` because `Arena` was responsible for deploying the contracts and thus needed the abis. It's no longer the case that `Arena` now isn't responsible for deploying the contracts.

permissionless-arbitration/contracts/src/CanonicalConstants.sol
GCdePaula, line 1: For a later PR, we could consider generating this file automatically from some configuration, but I'm not sure.

permissionless-arbitration/contracts/test/MultiTournament.t.sol
GCdePaula, line 1: It would be nice have different constants for testing. Let's think more on this later!

GCdePaula (commented):
Looking good! I'm still reviewing.

offchain/core/src/arena/sender.rs
```diff
@@ -0,0 +1,285 @@
...
+            wallet.with_chain_id(config.web3_chain_id),
+        ));
+
+        Ok(Self { client })
+    }
+}
+
+/// The [ArenaSender] trait defines the interface for the creation and management of tournaments.
+#[async_trait]
+pub trait ArenaSender: Send + Sync {
```
GCdePaula, line 43: If you want to get rid of `Box<dyn Error>`, we can create an associated type:

```
trait T {
    type Error;

    async fn foo(
         &self,
    ) -> Result<(), Self::Error>;
}

struct S {}

impl T for S {
    type Error = MyConcreteError;
    // ...
}
```

GCdePaula (approved):
This looks great to me! I've added a few comments that, if you agree with, we can create issues for later.

PR #28, opened 2024-03-07
Embed Cartesi Machine

GCdePaula:
Attempt at embedding `libcartesi.a` in `cartesi-machine-sys`. There's a feature flag for linking with `libcartesi_jsonrpc.a` instead.

This is using an unmerged branch of the emulator that has the ongoing output unification.


GCdePaula:
I'll need help testing, documenting and general cleanups. Please give all the feedbacks!

stephenctw:
This is looking great! I did something similar and smaller on my local branch so I can run the tests with newer emulator version. But I used emulator `v0.16.0` instead of the `output-unification` branch. Why did you use the unmerged branch? I noticed the `v0.16.0` is somehow incompatible with `output-unification` when I ran both the simple and stress tests. Also another concern is that we have a few PRs pending right now, some of them are modifying the same code. We might want to prioritize merging them in proper order to avoid greater trouble in the future.

GCdePaula:
Hey Stephen! Thanks :)

> Why did you use the unmerged branch? 

This branch (Marcelo's) has the output unification that we'll need for Dave rollups. That's the main reason.

On the tests, I think it may have something to do changes in the emulator itself. I expect that that images we're testing with are not compatible. Just a guess, I'll investigate it!

> We might want to prioritize merging them in proper order

You're absolutely right.

algebraic-sofia (approved):
I think it's working correctly! The build is taking a little bit of time but I guess it's the intention of the crate. I think that if the machine emulator repository has tests we could just run them in the `cargo test`  command.

stephenctw:
> This branch (Marcelo's) has the output unification that we'll need for Dave rollups. That's the main reason.
I can rebase all my current work on this branch, but it's annoying that the tests are broken.

GCdePaula:
> I can rebase all my current work on this branch

You're using the jsonRPC bindings, right? I don't think this affects you.

In any case, I think we should merge your work to main first, then rebase this PR on top of main.

stephenctw:
> > I can rebase all my current work on this branch
> 
> You're using the jsonRPC bindings, right? I don't think this affects you.
> 
> In any case, I think we should merge your work to main first, then rebase this PR on top of main.

I want to use `machine-bindings` for new emulator release so I can work on the `reset` feature. jsonRPC is outdated and I thought we agreed to discard it?

GCdePaula:
True, you need the bindings for the reset.

The compute branch is currently using the jsonRPC bindings? Have you already started the reset?

I think these Rust bindings Sofia did were already at Marcelo's branch (maybe not the most recent commit).

We could make this binding for the `0.16` release of the machine. Not sure if there are many changes to the interface itself. Maybe they renamed rollups to cmio (Cartesi Machine IO).

If the changes are small, it's possible just checking out the right emulator branch can work. Otherwise, we may have to do some changes, but I think a simple regex can work.

We could ask Fusco to generate a docker image at that branch for us too.

I think the issue is that the CM images we have were built by an incompatible emulator version. In that case, if we want to use the bindings at Marcelo's branch, we'd have to generate those CM images again with the emulator at his branch.

But another possibility is that the bindings have some bug.

Also, with this PR, I think we can run the machine without docker. We may only need docker to create CM images.

This comment is a bit jumbled, I have a plane to catch soon. Let me know what you think, I'm not sure what's the best path forward.

stephenctw:
I switched to use the Rust bindings Sofia made on my local compute branch and made some small fixes so now things are working with `emulator v0.16.0`. My concern is that the binding now can conflict because of the changes you made and the changes I made were based on different tag/branch. I can push it to remote so you can take a closer look.

stephenctw:
> Ok! Your work is the priority. Let's finish it first, then worry about this PR and sort out the conflicts.
> 
> You don't need anything from this PR, right?

For the `rust-base` and `rust-compute` PR, I don't need anything for this PR. For the `uarch-reset` PR, I can use the embedded machine crate, but I can deal with it later.

stephenctw:
This looks great! I just have some questions about the `machine-runner`.

rollups-node/machine-runner/src/lib.rs
```diff
@@ -1,66 +1,158 @@
...
 
-    fn advance(&mut self, data: &[u8]) -> Result<()> {
-        self.feed(data)?;
+impl MachineRunner {
+    fn process_rollup(&mut self) -> Result<()> {
+        loop {
+            self.advance_epoch()?;
+
+            if self.epoch_number == self.state_manager.epoch()? {
+                break Ok(());
```
stephenctw, line 55: Can you elaborate more on this `break Ok(());` and its condition?

rollups-node/machine-runner/src/lib.rs
```diff
@@ -1,66 +1,158 @@
...
+        loop {
+            match self
+                .state_manager
+                .input(self.epoch_number, self.next_input_index_in_epoch)?
+            {
+                Some(input) => {
+                    self.process_input(&input)?;
+                    self.next_input_index_in_epoch += 1;
+                }
+                None => {
```
stephenctw, line 73: Is it possible that there are more inputs to be processed, but just not available by the time of the query?

rollups-node/machine-runner/src/lib.rs
GCdePaula: The `state_manager.epoch()` should return the latest epoch (accumulating) on the blockchain. We could improve this name probably. Maybe `consensus_accumulating_epoch`?

This function first process all inputs that currently exist in `self.epoch_number`, through the             `self.advance_epoch()` call. After processing these currently existing inputs, it checks the `consensus_accumulating_epoch`.

If the `consensus_accumulating_epoch` is "ahead", it means that the epoch with number `self.epoch_number` was sealed; the "currently existing inputs" are all the inputs that will ever exist for that epoch. This condition is the "else" block: the code will process the computation hash (TODO), increment the epoch and reset the input number, and loop (start processing the next epoch).

If the `consensus_accumulating_epoch` is the same as `self.epoch_number`, it means that the epoch with number `self.epoch_number` is not sealed; the "currently existing inputs" list is (possibly) incomplete.  There might be more inputs for this epoch.

rollups-node/machine-runner/src/lib.rs
GCdePaula: Yes!

rollups-node/machine-runner/src/lib.rs
stephenctw: So I think this `break Ok(());` should be `continue` instead? Therefore in the next iteration, it'll try to retrieve next input from current epoch again.

rollups-node/machine-runner/src/lib.rs
stephenctw: I personally would like move the `epoch` check from `process_rollup` into `advance_epoch`. So `advance_epoch` will only return if it completely processes all inputs from the current epoch, or in case of an error happen. I feel that it's more consistent with the function name this way. (just personal thought)

rollups-node/machine-runner/src/lib.rs
GCdePaula: Like, the `process_rollup` is like a "tick". It advances as much as it possibly can at that moment and returns. If we call it again, it should avance with newly available data. So we need to "poll" it, such as in a call/sleep/loop cycle.

stephenctw: Got it! Would you like to merge this PR, or is there something else you still working on?

GCdePaula: I wanted to merge it.

stephenctw (approved):

PR #37, opened 2024-04-09
Add `InMemoryStateManager`

stephenctw:


offchain/input-manager/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn get_index(&self) -> u64;
+}
+
+pub trait InputManager {
+    fn add_input(&mut self, input: Bytes, index: u64) -> Result<()>;
+    fn get_input(&self, index: u64) -> Option<&Bytes>;
+}
+
+pub struct InMemoryInputManager {
+    inputs: Vec<Bytes>,
```
GCdePaula, line 17: I think we can try to not use anything from ethers here. I think this could be a `Vec<u8>` of binary blob (not a hex string).

offchain/input-manager/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use anyhow::Result;
+use ethers::types::Bytes;
+
+pub trait IndexedInput {
+    fn get_index(&self) -> u64;
+}
+
+pub trait InputManager {
+    fn add_input(&mut self, input: Bytes, index: u64) -> Result<()>;
```
GCdePaula, line 12: Changing Bytes to `Vec<u8>`, we could maybe change this argument to `&[u8]`, or actual `Vec<u8>`. We just have to think about move semantics and try to minimize clones/copying. I have the feeling `&[u8]` is ideal, the caller borrows the blob, and the callee copies it, which I think is inevitable when we use SQLite.

offchain/input-manager/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn get_input(&self, index: u64) -> Option<&Bytes>;
```
GCdePaula, line 13: Rust style recommends not to use `fn get_x()` naming, preferring just `fn x()`.

offchain/input-manager/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+use anyhow::Result;
+use ethers::types::Bytes;
+
+pub trait IndexedInput {
+    fn get_index(&self) -> u64;
+}
+
+pub trait InputManager {
```
GCdePaula, line 11: My initial idea was to have this object be a general state manager, rather than just inputs. We'll need to store machine state hashes too, claims, etc. I was thinking we could make the same object handle everything, encapsulating a database connection/handle.

offchain/input-reader/src/lib.rs
```diff
@@ -48,12 +46,12 @@ impl InputReader {
...
                 provider: Provider::<Http>::try_from(provider_url)?,
                 semaphore: Semaphore::new(concurrency_opt.unwrap_or_default()),
-                app
+                app,
             },
         })
     }
 
-    pub async fn next(&mut self) -> Result<Vec<InputAddedFilter>, Box<dyn std::error::Error>> {
+    pub async fn next<E: EthEvent>(&mut self) -> Result<Vec<E>, Box<dyn std::error::Error>> {
```
GCdePaula, line 54: I like the use of generic. Another way to use generics is, instead of making them a parametric type of all these functions, adding them as a parametric type of the `InputReader`. You'd probably have to use something like [this](https://doc.rust-lang.org/rust-by-example/generics/phantom.html). Just a suggestions.

The difference is that you'd need to use the "turbo fish" syntax when instantiating `InputReader`, and never again. This design is possible because we expect `next` will always be called with the same event type.

It could make sense to change `InputReader` to have parametric type on the struct and not on the function `next`, but still keep the functions with parametric types in `PartitionProvider`. If `PartitionProvider` was shared, it's possible someone would want to read different event types. I don't think we should make it shared, this is great.

offchain/input-reader/src/lib.rs
```diff
@@ -5,17 +5,15 @@ use async_recursion::async_recursion;
...
 use ethers::prelude::{Http, ProviderError};
 use ethers::providers::{Middleware, Provider};
 use ethers::types::{Address, BlockNumber, Filter, H160, U64};
-
 struct PartitionProvider {
     provider: Provider<Http>,
     semaphore: Semaphore,
     input_box: Address,
-    app: H160
+    app: H160,
```
GCdePaula, line 16: I think I'd move `input_box` and `app` to `InputReader`, and pass it as an argument.

offchain/Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "dave-advance",
     "input-query",
+    "input-manager",
```
GCdePaula, line 6: Let's also think about whether things should be crates or if just modules are enough. I think this is fine the way it is, but sometimes a module is good enough.

offchain/input-reader/src/lib.rs
```diff
@@ -48,12 +46,12 @@ impl InputReader {
                 input_box,
                 provider: Provider::<Http>::try_from(provider_url)?,
                 semaphore: Semaphore::new(concurrency_opt.unwrap_or_default()),
```
GCdePaula, line 48: A builder pattern might be interesting for all the optional parameters.

GCdePaula (commented):
Looking good!

GCdePaula (approved):
LGTM!

PR #39, opened 2024-04-25
Add doom showcase

stephenctw:
- code cleanup
- download machine templates from git release (should we use IPFS instead?)
- support `DEPLOY_TO_ANVIL` option through env for lua scripts
- support `CHAIN_ID`, `PRIV_KEY`, `ROOT_TOURNAMENT`, `URL` options through env for `dave-compute`


permissionless-arbitration/lua_node/utils/helper.lua
guidanoli, line 1: Why not declare the module table in the beginning...

```lua
local helper = {}
```

... and, for each function, index it?

```lua
function helper.myfoo()
  -- [ ... ]
end
```

guidanoli:
- [x] Build test image
- [x] Run simple test
- [ ] Run stress test :question:
- [x] Run Doom showcase

I don't know if the stress test was successful or not. It just prints out `Hello from Dave lua prototype!` and exits with status code 137, which tells me it might not have been successful?

Also, I felt the timeout quite dreadful. Would it be possible to reduce it just for the purposes of running the tests locally and not having to wait such a long time? I might not be aware of any undesired side effects of such a change, so please tell me if there is a reason why we shouldn't mess with the timeout just for testing purposes.

stephenctw:
@guidanoli It's definitely not successful with exit code 137. Is there any more information from the console logs? Also, which version did this happen? The `lua-node` version or the `offchain` + `lua-node` combination? For the timeouts, I think most of the time are spent on the machine computation. But the timing constants are set based on my machine performance, it may vary from yours. I'll try to update a measurement script that @GCdePaula wrote and you can try to run it to generate different timing constants for your machine.

stephenctw:
@guidanoli now you can go to `permissionless-arbitration/measure_constants` directory and run the script to generate better fitting timing constants for your machine.

guidanoli:
> Is there any more information from the console logs? 

No, just the line `Hello from Dave lua prototype!`.

> which version did this happen?

I just ran the command that was on the README under `permissionless-arbitration/lua_node`:

```
docker run --rm \
    --env MACHINE_PATH="/app/lua_node/program/debootstrap-machine-sparsed" \
    --env DEPLOY_TO_ANVIL="true" \
    dave:latest
```

permissionless-arbitration/measure_constants/README.md
```diff
@@ -0,0 +1,33 @@
...
+Valid value for `MACHINE_PATH` can be:
+
+-   `simple-program`
+-   `doom-compute-machine`
+-   `debootstrap-machine-sparsed`
+
+There are two values in the script that can be configured (WIP):
+
+-   `root_tournament_slowdown`:
+    This is what the user (you) finds acceptable as slowdown for the root tournament. In other words, whichever stride the script choses for the root tournament, calculating the commitment cannot slow the machine more than `root_tournament_slowdown` when compared with just calculating the final state. Default it to 2.5 slowdown with the reference machine. For rollups it can be 5.0.
```
guidanoli, line 16: chooses

guidanoli:
> now you can go to permissionless-arbitration/measure_constants directory and run the script to generate better fitting timing constants for your machine.

Thanks! Tomorrow I will try it out.

guidanoli:
- [x] Build test image
- [x] Run simple test
- [x] Run stress test
- [x] Run Doom showcase

@stephenctw and I hopped on a call to investigate the issue with the stress test not working, and it seems that the issue was the memory limit set by Docker. I set the memory limit to 3.8 GB, and swap to 2 GB, and it worked like a charm. Thanks, Stephen!

stephenctw:
@guidanoli I add the graphic feature to the doom showcase. I only add this to the Lua prototype as it can reduce performance of the Dave process. Please let me know if I should also add this to the Rust Dave (production code).

I did this in a hacky way, suggestions are very welcomed.

- Dave Lua node snapshot the machine at the beginning of every level of tournament, also store extra information about the current commitment tree.
- A separate script monitors the snapshot directory constantly. When a new snapshot is taken, it implies the node has reached a new level of tournament. Therefore, the script loads the snapshot, run the machine for the given range of cycles (calculated from the extra information in the snapshot), and then read the game frames, process them and save to png files.
- The user runs another script on his host machine (`./doom-showcase/view.sh`) that will play the frames repeatedly as the png files are being generated. The game play starts fragmental and stops when the full game play is showed (of the current level).
- In reality, you'll see the same video throughout the four levels of tournaments. That's because the game play is short, and the logic of our dishonest player is designed to be simple. The dispute always goes on to the left most leaf, and thus the game play always starts from the beginning.

guidanoli:
Managed to run it!
The color palette is indeed a bit glitchy.
Some white and blue pixels seem out-of-place.

[Screencast from 01-05-2024 09:25:36.webm](https://github.com/cartesi/dave/assets/9125226/c4ec77ae-db41-4103-84e6-dc2049957b58)

guidanoli:
Maybe worth trying to extract the palette from the machine as feeding it into the `playpal2rgb` tool (which I updated to accept such palette as input rather than having it hard-coded).

stephenctw:
> Maybe worth trying to extract the palette from the machine as feeding it into the `playpal2rgb` tool (which I updated to accept such palette as input rather than having it hard-coded).

Sure, once Eduardo gave me the range of memory region, I'll update it.

stephenctw:
> In my case, I had to create the snapshot folder manually. Maybe worth putting instructions on the README to do so as well.
> 
> ```shell
> mkdir snapshots
> ```

I'll just add this line to the entrypoint script.

permissionless-arbitration/lua_node/doom_showcase/view.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
stephenctw: I think this `view.sh` script can fail sometimes but still being considered a normal behavior. For example, when one level tournament is done replaying the video, but the next level of tournament is yet preparing the frames. In such case we don't want the script to exit, but to keep trying until there's something to play for the next level. But this is a good suggestion, I'll apply to the other scripts in the project.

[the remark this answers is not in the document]
stephenctw: Actually, I should change the way the script behaves, and still apply your suggestion!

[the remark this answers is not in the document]
guidanoli: Thanks!

permissionless-arbitration/lua_node/doom_showcase/view.sh
```diff
@@ -1,22 +1,29 @@
...
+            if [ -d "done" ]; then
                 mplayer -mf fps=30 -x 640 -y 400 -geometry 50%:50% 'mf://*.png'
+                break
             else
                 mplayer -mf fps=30 -x 640 -y 400 -geometry 50%:50% 'mf://*.png'
-                break
             fi
         done
+        log2_stride=($( cat ../log2_stride ))
+        if [ "$str" == "0" ];then
```
guidanoli, line 23: Sorry, where is this `str` variable declared?

stephenctw: Ah that's a terrible typo

stephenctw:
Rebased the branch on top of main, updated all the examples to use `machine-emulator 0.17`.

GCdePaula (approved):
LGTM!

PR #42, opened 2024-05-20
Dave Rollups Node

stephenctw:
Based on design diagram: https://miro.com/app/board/uXjVKKcSTdw=/


stephenctw:
@GCdePaula I figured it's not too hard to use `sqlite` in the `StateManager` (with the work Sofia has already built earlier), so I abandoned the `InMemory` version of it. Does the schema make sense to you? There're still some parts missing but I think we can fill in the blanks later as we keep working toward the node. What do you think?

GCdePaula:
Great work! It's progressing quite nicely :)

Some comments on the SQL:

#### 1. Constants Table:

This is the only table I'm unsure about. Mainly because of this weak/loose typing. Let's discuss this one more carefully. In the diagram, we have:

```
Machine initial state snapshot
Block created
Contact Addresses
```

The machine initial state snapshot, could it go to the snapshot table maybe? I don't know. The contact addresses, does it make sense to have a dedicated table, like address book? Also, these configurations are added when the process starts, probably in a json/toml. Now I'm thinking if we need to replicate it at all. But I also like the idea of having them in table.


#### 2. Epochs Table:

I think the `sealed` should also be `NOT NULL`. Like, this table should only contain epochs that have a boundary. Also, I think we can use longer names, like `block_sealed` or `end_block`.

#### 3. Inputs Table:

The input index is the input index _in epoch_. I think we should add that to the name to be clear, and think about how we'll get that index in the `blockchain_reader`.

#### 4. States Table:

I'd extend the name of this table to include something like "state", "hashes", "machine", "execution". I don't know, `machine_state_hashes` or `execution_state_hashes` or just `state_hashes`. Also rename the column `state` to `state_hash` or something. Also, the input index is also in the epoch.

---

I've never used SQL much. Are there things where we tell SQLite that some indices are monotonic? Intuitively, it should speed up the queries. Tables where there are two primary keys which are indices, there might be a way to tell SQLite there's an order, by sorting first by one key and then by the other?

stephenctw:
I agree that the `Constants Table` can be well supplied from the command line arguments or a configuration file. There's also a value that may not fit into any of the above tables, the `latest_processed_block`. I think I'm also lost about how we should use `meta-counter`.

GCdePaula:
We can change the metacounter field of the snapshot to just `epoch_number` and `input_number`. To recover the metacounter from `input_number` we just do `input_number << (a+b)`.

The metacounter contains info on "where we are": which input, which big instruction, which micro-instruction. This machine runner won't touch the uarch, so all those bits are zero. To simplify, we should also only take snapshots after a whole inout has been processed, so all those bits are also zero.

Thinking now, we could even drop the metacounter on this part of the code, it's truly only needed in the dispute resolution.

GCdePaula:
I think we should skip the full state snapshots for now and leave it for later.

GCdePaula:
I think we can delete `- settled: BOOLEAN, NOT NULL` column too. Not sure.

stephenctw:
I think I need some help on the `machine` front. I'm not sure if the break/yield reasons are handled properly.

GCdePaula:
Ok! I've been working on PR #28, and I think I might have changed how this is handled. I'll clean it up and push it.

rollups-node/state-manager/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = { workspace = true }
+edition = { workspace = true }
+homepage = { workspace = true }
+license-file = { workspace = true }
+readme = { workspace = true }
+repository = { workspace = true }
+
+[dependencies]
+anyhow = "1.0.6"
+sqlite = "0.33.0"
```
GCdePaula, line 15: I think we should use this crate instead: https://github.com/rusqlite/rusqlite

rollups-node/state-manager/src/lib.rs
```diff
@@ -0,0 +1,486 @@
...
+use sqlite::State;
+
+pub struct StateManager {
+    connection: sqlite::ConnectionThreadSafe,
+}
+
+impl StateManager {
+    pub fn connect(database_uri: &str) -> Result<Self> {
+        let connection = sqlite::Connection::open_thread_safe(database_uri)?;
+        // constants table
```
GCdePaula, line 14: I'm not good with SQL in general. Was there a concept of "migrations" to deal with table creation?

GCdePaula: https://cj.rs/rusqlite_migration/

prt/compute-rs/core/src/contract/leaf_tournament.rs
GCdePaula, line 1: I think we shouldn't commit this and other generated contract binding files.

GCdePaula (commented):
Looks good, just one comment!

prt/compute-rs/core/src/contract/leaf_tournament.rs
stephenctw: I'll remove them in a separate [PR ](https://github.com/cartesi/dave/pull/53)to avoid conflicts.

GCdePaula (approved):
🚀🚀🚀

PR #44, opened 2024-05-31
Feature/dave rollups stephen

stephenctw:


stephenctw:
@GCdePaula , I modified the `blockchain-reader` so it can retrieve both `input` and `epoch` events. I don't particularly like the way I did it, do you have any suggestions on how it can be further improved?
Also, does the `start` method of `machine-runner` looks like something you had in mind?

rollups-node/machine-runner/src/lib.rs
```diff
@@ -37,25 +40,34 @@ impl MachineRunner {
...
+        // process all inputs that are currently availalble
         loop {
             self.advance_epoch()?;
+            let latest_epoch = self.state_manager.epoch()?;
 
-            if self.epoch_number == self.state_manager.epoch()? {
+            if self.epoch_number == latest_epoch {
                 break Ok(());
-            } else {
+            } else if self.epoch_number < latest_epoch {
```
GCdePaula, line 63: I think this should be an assert instead:

```rust
else {
  assert(self.epoch_number < latest_epoch);
  // ...
}
```

Also, the body of this else is where we do the computation hash. Let's add a TODO comment here to calculate computation hash and save it to database.

rollups-node/blockchain-reader/src/lib.rs
```diff
@@ -3,117 +3,158 @@
...
-    __phantom: std::marker::PhantomData<E>,
+    sleep_duration: Duration,
 }
 
-impl<E: EthEvent> BlockchainReader<E> {
+impl BlockChainReader {
     pub fn new(
-        app: Address,
+        app_genesis: u64,
+        consensus: Address,
```
GCdePaula, line 28: I think we could create an `AddressBook` type that has all these addresses (app, input, consensus).

GCdePaula (commented):
It's looking good!

Let me think more about the `start`, there are annoying details like shutting down, etc, but conceptually it looks ok to me. I think there are some other ways to structure this start, such as having `new` be a Builder style, and the `start` function consumes it (_i.e._ takes `self` instead of `&self`). But the overall logic won't be that much different.

stephenctw:
@GCdePaula , should we change the `epochs` table to include a nullable `claim/commitment` column?
```
CREATE TABLE epochs (
    epoch_number INTEGER NOT NULL PRIMARY KEY,
    last_input_index INTEGER NOT NULL,
    claim BLOB
);
```
So, the machine-runner will calculate the commitment when an epoch is sealed and update it to the database.

GCdePaula:
I think the commitment should be a separate table. This way, all our rows are immutable, and the tables "append only" (except the last processed blocks I guess).

stephenctw:
> I think the commitment should be a separate table. This way, all our rows are immutable, and the tables "append only" (except the last processed blocks I guess).

Something like this?
```
CREATE TABLE commitments (
    epoch_number INTEGER NOT NULL PRIMARY KEY,
    claim NOT NULL BLOB
);
```

BTW, should we call it commitment or claim?

rollups-node/state-manager/src/sql/migrations.sql
```diff
@@ -19,13 +19,15 @@ INSERT INTO latest_processed (id, block) VALUES (1, 0);
 CREATE TABLE machine_state_hashes (
     epoch_number INTEGER NOT NULL,
     input_index_in_epoch INTEGER NOT NULL,
+    state_index_in_input INTEGER NOT NULL,
+    repetitions INTEGER NOT NULL,
     machine_state_hash BLOB NOT NULL,
-    PRIMARY KEY (epoch_number, input_index_in_epoch)
+    PRIMARY KEY (epoch_number, input_index_in_epoch, state_index_in_input)
 );
```
GCdePaula, line 26: These fields I believe should be on a different table.

GCdePaula:
@stephenctw Let's call it `computation_hashes` and `computation_hash`!

rollups-node/state-manager/src/sql/migrations.sql
GCdePaula: Your suggestion of:

```
CREATE TABLE commitments (
    epoch_number INTEGER NOT NULL PRIMARY KEY,
    claim NOT NULL BLOB
);
```

makes sense. 

I *think* this one should be just:

```
CREATE TABLE machine_state_hashes (
    epoch_number INTEGER NOT NULL,
    index INTEGER NOT NULL,
    repetitions INTEGER NOT NULL,
    PRIMARY KEY (epoch_number, index)
);
```

rollups-node/state-manager/src/persistent_state_access.rs
```diff
@@ -1,20 +1,23 @@
...
+use crate::{Epoch, Input, InputId, StateManager};
 
-use rusqlite::Connection;
+use rusqlite::{Connection, OptionalExtension};
+use std::sync::Mutex;
 
+#[derive(Debug)]
 pub struct PersistentStateAccess {
-    connection: Connection,
+    connection: Mutex<Connection>,
```
GCdePaula, line 12: Why do we need a mutex here?

GCdePaula: I'm reading, I think we can create multiple connections.

GCdePaula: Yeah, I think it is possible for each thread to have its own connection. Can you read up on this too? I think there are settings like WAL that help with contention etc, but it seems that the default settings already allow for this.

stephenctw: > Yeah, I think it is possible for each thread to have its own connection. Can you read up on this too? I think there are settings like WAL that help with contention etc, but it seems that the default settings already allow for this.

Ok, I'll remove the `Mutex` and create separate connections for each thread in the `spawn` call.

stephenctw: So, I gave it a try and got yet another similar error:
```
error[E0277]: `RefCell<rusqlite::inner_connection::InnerConnection>` cannot be shared between threads safely
   --> dave-rollups/src/lib.rs:38:5
    |
38  | /     spawn(async move {
39  | |         let state_manager = create_state_manager(&params.sqlite_connection_string).unwrap();
40  | |         let mut blockchain_reader = BlockchainReader::new(
41  | |             state_manager,
...   |
53  | |             .unwrap();
54  | |     })
    | |______^ `RefCell<rusqlite::inner_connection::InnerConnection>` cannot be shared between threads safely
    |
    = help: within `BlockchainReader<PersistentStateAccess>`, the trait `Sync` is not implemented for `RefCell<rusqlite::inner_connection::InnerConnection>`, which is required by `{async block@dave-rollups/src/lib.rs:38:11: 54:6}: Send`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock` instead

```
And also found that `rusqlite` isn't working well with `async`: https://github.com/rusqlite/rusqlite/issues/1013

stephenctw: There are other projects that are built and based on the original `rusqlite` that provide the `tokio/async` feature:
https://github.com/programatik29/tokio-rusqlite
https://github.com/jsdw/async-rusqlite

I think for now we could move on with the `Mutex` solution, create an issue and come back later to decide which is the best way to go. What do you think?

GCdePaula: Ok, good idea!

GCdePaula (approved):

PR #47, opened 2024-06-19
chore: refactor contract interfaces

stephenctw:
- split tournament factory into single/multi level
- update default top tournament for `lua_poc` and `dave-compute`
- update rust bindings of prt contracts


GCdePaula:
Hey @stephenctw!

There was some similar work here #49. These changes were to enable @guidanoli to start his own work. The changes on the bindings were missing in #49. Also, do we need to commit those bindings?

I think we should start merging some stuff. How do you think we should do that?

GCdePaula:
Ok!

GCdePaula:
Ok! Sorry for the confusion.

PR #54, opened 2024-07-27
Enrich dishonest node behavior

stephenctw:
- allow multiple fake commitment from one dishonest player
- implement `win_match_by_timeout` in offchain (both Lua and Rust nodes)
- bug fixes
- general code improvements


GCdePaula (approved):
LGTM

PR #56, opened 2024-07-29
correct `running_leaf` for `seal_match` offchain

stephenctw:
This is to match the `setDivergence` logic onchain
https://github.com/cartesi/dave/blob/main/prt/contracts/src/Match.sol#L284-L305


stephenctw:
Also improve the json output files for easier adoption.

GCdePaula (approved):

PR #63, opened 2024-08-13
Refactor bindings generation

stephenctw:


prt/Makefile
```diff
@@ -0,0 +1,29 @@
...
+	@echo '  clean                      - clean the cache artifacts and generated files'
+	@echo '  bind                       - generate Rust bindings from Solidity code'
+
+clean:
+	@$(CLEAN_CMD)
+	@$(FORGE_CMD) clean
+
+bind: clean
+	@$(FORGE_CMD) bind --ethers && \
+	for file in $(BINDING_FILES); do \
```
GCdePaula, line 25: I think there's an option on the `forge bind` command to do this. Take a look at all the options, it's quite complete!

prt/contract-bindings/build.rs
GCdePaula, line 1: I'm not sure about this file. Should we have a build script, or just the makefile? If we add the build script, we're adding forge as a dependency.

GCdePaula (commented):
Looking good!

prt/Makefile
```diff
@@ -0,0 +1,21 @@
...
+	@echo '  clean                      - clean the cache artifacts and generated files'
+	@echo '  bind                       - generate Rust bindings from Solidity code'
+
+clean:
+	@$(CLEAN_CMD)
+	@$(FORGE_CMD) clean
+
+bind: clean
+	@$(FORGE_CMD) bind --ethers --select $(BINDINGS_FILTER) && \
+	cp -r ./out/bindings/src ../$(BINDINGS_DIR)
```
GCdePaula, line 19: There's the `-o` or `--out` flag for this!

prt/Makefile
```diff
@@ -0,0 +1,21 @@
+BINDINGS_DIR := ./contract-bindings
+SRC_DIR := ./contracts
+BINDINGS_FILTER := 'LeafTournament|RootTournament|^Tournament$$'
+
+# Common commands
+CLEAN_CMD := find . -wholename '$(BINDINGS_DIR)/*.rs' -delete
```
GCdePaula, line 6: There's the `--overwrite` flag, which I think may substitute this delete phase. We probably also want the `--skip-cargo-toml` flag, since we're providing our own.

prt/Makefile
```diff
@@ -0,0 +1,21 @@
...
+FORGE_CMD := cd $(SRC_DIR) && forge
```
GCdePaula, line 7: There's probably a flag to remove the need of this `cd`. Maybe `--bindings-path`? Not sure.

prt/Makefile
GCdePaula: Another possibility is the `--module` flag. We tell forge to generate a module and write it _inside_ our crate; this way we don't need to worry about forge overwriting our crate files (like `Cargo.toml` or `lib.rs`).

In this approach, our `lib.rs` would just include/reexport this module.

GCdePaula (commented):
Some ideas! I'm not sure they work.

prt/Makefile
stephenctw: The second approach seems nicer.

GCdePaula (commented):
Nice! Can you verify whether we need a `.gitkeep` inside `src/contract`? Also, I think a clean command that deletes the contents of `src/contract` might make sense, for the user to call if they want.

stephenctw:
> Nice! Can you verify whether we need a `.gitkeep` inside `src/contract`? Also, I think a clean command that deletes the contents of `src/contract` might make sense, for the user to call if they want.

`forge` is a little bit picky about the output binding directory. The output binding directory either has to not exist at all, or it has to be called with a `--overwrite` option. I suggest not to commit `src/contract` directory, thus no `.gitkeep` is necessary. So the crate would contain `src/lib.rs`, `Cargo.toml` and that's it. `src/contract` module will be created directory from the `make bind` command.

And yes, I'll add the clean action to Makefile.

GCdePaula (approved):
✨✨

PR #64, opened 2024-08-14
Refactor merkle

GCdePaula:


stephenctw:
I see there're two `Digest` implementation (and they look very similar), one under `merkle`, the other one under `cartesi-machine`. Do you think we should merge them into one?

GCdePaula:
Good catch! I hadn't realized that.

But thinking more, I believe we need both. The digest in merkle is about commitments, and the digest in the machine are about its state. Concretely, we could have the digest of the machine be smaller, using a different hashing function. I think we shouldn't merge them yet.

stephenctw (approved):

PR #65, opened 2024-08-28
Replace `ethers` with `alloy`

stephenctw:
This turns out to be a much bigger PR than just replacing `ethers` with `alloy`.

- full migration to `alloy` from `ethers`
- update `machine-bindings` build script for `emulator v0.18`
- full `node` using `DaveConsensus` and `rollups-contracts` (still missing the part spawning a compute dispute process)
- all compute tests passed with both `client-lua` and `prt-rs`


GCdePaula (approved):
🔥🔥🔥

PR #71, opened 2024-09-06
feat: add view methods for other immutable variables

guidanoli:


guidanoli:
The `IInputBox` variable is going to be made immutable in #70, but this can be merged before.

GCdePaula (approved):
Should you just mark the variables as public?

PR #76, opened 2024-09-15
Apply scope and coroutines to compute tests

stephenctw:
Each Lua player is wrapped in a coroutine with isolated scope.


prt/tests/compute/runners/helpers/fake_commitment.lua
```diff
@@ -1,7 +1,7 @@
 local consts = require "computation.constants"
 local MerkleBuilder = require "cryptography.merkle_builder"
 local Hash = require "cryptography.hash"
-local CommitmentBuilder = require "computation.commitment"
+local new_scoped_require = require "utils.scoped_require"
```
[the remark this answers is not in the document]
GCdePaula: One way to implement these fake players is by calculating the honest commitment once, and then doing splices on it.

For example, suppose you know the honest commitment. The way change only the final state is the following (not tested!!):

```
local function change_last(root, new_last)
  local left, right = root:children()
  if not right then
    return new_last
  else
    return keccak(left, change_last(right, new_last))
  end
end
```

The above function receives a commitment and returns a new commitment, with only the last state different. Here's a way to make it arbitrary (not tested!!):

```
local function change_leaf(root, height, new_leaf, index)
  if height == 0 then
    return new_leaf
  else
    local left, right = root:children()
    assert(left); assert(right)
    if (index >> (height-1)) & 1 == 0
      return keccak(change_leaf(left, height-1, new_leaf, index), right)
    else
      return keccak(left, change_leaf(right, height-1, new_leaf, index))
    end
  end
end
```

prt/tests/compute/runners/helpers/fake_commitment.lua
```diff
@@ -15,69 +15,92 @@ local function build_zero_uarch()
...
-    -- local new_internalized_hashes = shallow_copy(internalized_hashes)
+local function build_commitment(machine_path, base_cycle, level, log2_stride, log2_stride_count)
+    -- the honest commitment builder should be operated in an isolated env
+    -- to avoid side effects to the strategy behavior
 
-    local c = builder:build(base_cycle, level, log2_stride, log2_stride_count)
-    -- restore the internal states of `Hash` from previous snapshot
-    -- this is to avoid the dishonest player to react on the honest behave
-    -- Hash.set_internal(new_internalized_hashes)
+    local c = coroutine.create(function(...)
```
GCdePaula, line 62: This function doesn't need to be variadic.

prt/tests/compute/runners/hero_runner.lua
```diff
@@ -2,46 +2,35 @@
 require "setup_path"
```
GCdePaula, line 2: I think this `require "setup_path"` may not be needed anymore. Can you test?

Also the `#!/usr/bin/lua` maybe?

prt/tests/compute/utils/scoped_require.lua
```diff
@@ -15,12 +38,23 @@ local function new_scoped_require(env)
 
     local function scoped_require(name)
         if loaded[name] == nil then
```
GCdePaula, line 40: Can you change this line to just `if not loaded[name] then`?

prt/tests/compute/utils/scoped_require.lua
GCdePaula, line 1: Great changes!!

GCdePaula (commented):
Looking good! Left a few comments.

prt/tests/compute/runners/helpers/fake_commitment.lua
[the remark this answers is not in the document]
stephenctw: > It's ok to sandbox the dishonest players from each other, but it might make sense to allow the dishonest players to "coordinate" and see each other's states.

This is still the case, I'm only creating separate scope to isolate the `honest commitment` calculation as part of the `fake commitment` construction to avoid the sybil players to know and act as if they are honest.

prt/tests/compute/runners/helpers/fake_commitment.lua
stephenctw: Good catch, I just accept auto-completion too quickly without checking.

prt/tests/compute/utils/scoped_require.lua
stephenctw: Yeah, otherwise the `cartesi` module cannot be loaded!

prt/tests/compute/prt_compute.lua
```diff
@@ -2,152 +2,110 @@
...
+        local status = coroutine.status(c)
 
-    if active_players > 0 then
-        if helper.all_players_idle(pid_player_map) then
-            all_idle_count = all_idle_count + 1
-            helper.rm_all_players_idle(pid_player_map)
-        else
-            all_idle_count = 0
+        if status == "dead" then
+            table.remove(player_coroutines, i)
```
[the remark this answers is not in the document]
stephenctw: Good catch. I think if I iterate the table in backwards, I get to keep the `table.remove` and the same logic for checking whether the program is finished.

prt/tests/compute/prt_compute.lua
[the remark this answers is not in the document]
stephenctw: This is interesting. We should also consider the case when we are running the Rust dispute node. It may just be easy to write to some file at a designated path, where all the temporary snapshots are stored for the current tournament.

prt/tests/compute/runners/helpers/fake_commitment.lua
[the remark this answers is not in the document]
stephenctw: Hey @GCdePaula , I think your algorithm is very efficient in the way that it only rebuilds the branch of the tree that contains the modified leaf and update the root hash accordingly. However things are a bit more tricky since the `honest commitment` is calculated in a separate scope now, so it becomes inevitable to rebuild the whole thing using the `Hash` in current `fake commitment` scope, to rebuild the children relationship. To apply your algorithm, the `honest commitment` must not be done in a separate scope, this may be doable since I added the extra check on the player's turn of the match. But I think you like it being isolated so it's how it should behave in realworld in production. We can schedule a call to discuss this if you think is necessary.

prt/tests/compute/blockchain/utils.lua
```diff
@@ -19,4 +19,10 @@ local function advance_time(seconds, endpoint)
     end
 end
 
-return { advance_time = advance_time }
+local deploy_cmd = [[sh -c "cd ../../contracts && ./deploy_anvil.sh"]]
+local function deploy_contracts()
+    local reader = io.popen(deploy_cmd)
```
guidanoli, line 24: After `reader:read()`, shouldn't `reader:close()` be called, and the first return value be checked for `true`, meaning the shell command exited with success?

prt/tests/compute/prt_compute.lua
```diff
@@ -2,152 +2,107 @@
...
-        -- close the reader and delete the reader entry when there's no more msg in the buffer
-        -- and the process has already ended
-        if message_out == 0 and helper.is_zombie(pid) then
-            helper.log_full(pid_player_map[pid], string.format("player process %s is dead", pid))
-            reader:close()
-            pid_reader_map[pid] = nil
-            pid_player_map[pid] = nil
+    local idle = true
+    for i = #player_coroutines, 1, -1 do
+        local c = player_coroutines[i]
```
guidanoli, line 83: Why not `for i, c in ipairs(player_coroutines) do ... end`?

prt/tests/compute/prt_compute.lua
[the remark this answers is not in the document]
guidanoli: Maybe we could just skip dead coroutines, and keep track of whether there is at least one live coroutine, in order to break from the `while true` loop, like:

```lua
while true do
  local has_live_couroutine = false
  for i, c in ipairs(player_coroutines) do
    -- ...
    if status == "dead"
      goto continue
    end
    has_live_couroutine = true
    -- ...
    ::continue::
  end
  if not has_live_couroutine then
    break
  end
  -- ...
end
```

prt/tests/compute/utils/scoped_require.lua
```diff
@@ -14,13 +37,24 @@ local function new_scoped_require(env)
...
 
     local function scoped_require(name)
-        if loaded[name] == nil then
-            local path = find_chunk(name)
+        if not loaded[name] then
+            local module_type, path = find_path(name)
+
+            local chunk, err, result
+            if module_type == "lua" then
+                chunk, err = loadfile(path, "t", new_env)
```
guidanoli, line 45: Isn't `"bt"` the default value for the `mode` parameter of the `loadfile` function?

prt/tests/compute/prt_compute.lua
```diff
@@ -2,152 +2,110 @@
...
-            pid_player_map[pid] = nil
+    local idle = true
+    local has_live_coroutine = false
+    for i, c in ipairs(player_coroutines) do
+        if c then
+            local success, ret = coroutine.resume(c)
+            local status = coroutine.status(c)
+
+            if status == "dead" then
+                player_coroutines[i] = nil
```
guidanoli, line 89: If you set some `player_coroutines[i]` to `nil`, then, next time it iterates over `player_coroutines`, it is not going to visit `j >= i`. I would suggest just skipping a coroutine if the coroutine is dead, and not fiddle with `player_coroutines`.

stephenctw: Good catch! What if I just set it to `false` instead of `nil`?

guidanoli: I think that would work, yes. :-)

stephenctw:
Also added Rust compute node as a coroutine. There are two versions:
- Rust compute node reacts once to the tournament and then exits
- Rust compute node reacts periodically to the tournament

The first version doesn't perform well because it has to re-compute all the commitments everytime the process starts. It can improve a lot once we are able to cache and restore the commitments.

GCdePaula (approved):
✨✨✨

PR #77, opened 2024-09-20
Merkle for `DaveConsensus`

guidanoli:
- Adapted `MerkleV2` for 32-byte words
- Updated `DaveConsensus` to use that
- Fixed a bug in `MerkleV2` for data of length in range $(0,8)$
- Simplified function that gets hash of word at index
- Optimized operation of "joining" with EVM assembly (copied from `LibMerkle32`)
- Simplified function that gets pristine node at certain height


cartesi-rollups/contracts/src/Merkle.sol
```diff
@@ -0,0 +1,143 @@
...
+
+library Merkle {
+    using Math for uint256;
+
+    /// @notice Compute the parent of two nodes.
+    /// @param leftNode The left node
+    /// @param rightNode The right node
+    /// @return parentNode The parent node
+    /// @dev Uses assembly for extra performance
+    function parent(bytes32 leftNode, bytes32 rightNode) internal pure returns (bytes32 parentNode) {
```
GCdePaula, line 18: I like the name `join` for this method, but this is ok too.

GCdePaula (commented):
Looking good! Did you measure gas in any way, to compare with the previous version?

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: Yeah, `join` is good as well.

guidanoli:
> Did you measure gas in any way, to compare with the previous version?

I haven't compared them gas-wise, but I believe that would an interesting analysis to do later.

cartesi-rollups/contracts/src/Math.sol
```diff
@@ -0,0 +1,107 @@
...
+    function getLog2Floor(uint256 x) internal pure returns (uint8) {
+        require(x != 0, "log of zero is undefined");
+
+        return uint8(255 - clz(x));
+    }
+
+    /// @notice count trailing zeros
+    /// @param x number you want the ctz of
+    /// @dev this a binary search implementation
+    function ctz(uint256 x) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
guidanoli: Interesting. I would love to know why this works.

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: Applied in 8b8833c.

cartesi-rollups/contracts/src/Math.sol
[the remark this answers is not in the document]
guidanoli: Applied in 8b8833c.

cartesi-rollups/contracts/src/Merkle.sol
```diff
@@ -0,0 +1,143 @@
...
+            return PristineMerkleTree.getNodeAtHeight(log2NumOfWordsInDrive);
+        }
+
+        uint256 numOfWordsInDrive = 1 << log2NumOfWordsInDrive;
+
+        require(data.length <= (numOfWordsInDrive << MerkleConstants.LOG2_WORD_SIZE), "Data larger than drive");
+
+        uint256 numOfWordsInData = data.length.ceilDivExp2(MerkleConstants.LOG2_WORD_SIZE);
+
+        bytes32[] memory stack = new bytes32[](2 + numOfWordsInData.getLog2Floor());
```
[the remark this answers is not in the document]
guidanoli: The performance implications are indeed not very relevant.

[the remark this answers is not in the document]
guidanoli: Applied in b64557d.

cartesi-rollups/contracts/src/Math.sol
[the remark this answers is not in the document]
guidanoli: Now that we're using `log2SizeOfDrive` for the stack depth, we don't need `getLog2Floor` nor `clz`, so we're left with just the binary search implementation of `ctz`. See b64557d.

cartesi-rollups/contracts/src/MerkleConstants.sol
```diff
@@ -0,0 +1,11 @@
+// (c) Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0 (see LICENSE)
+
+pragma solidity ^0.8.0;
+
+library MerkleConstants {
+    uint256 constant HASH_SIZE = 32;
+    uint256 constant LOG2_MEMORY_SIZE = 64;
+    uint256 constant LOG2_WORD_SIZE = 5;
```
GCdePaula, line 9: Does the change in the machine state Merkle tree affects this?

GCdePaula: I think so, right? The leafs were just changed to be 32-bytes sized, instead of 8-bytes sized.

cartesi-rollups/contracts/src/Merkle.sol
```diff
@@ -0,0 +1,142 @@
...
+    /// @notice Get the Merkle root hash of a drive with a replacement.
+    /// @param position position of replacement in drive
+    /// @param log2SizeOfReplacement log2 of the size of the replacement
+    /// @param log2SizeOfDrive log2 of the size of the drive
+    /// @param replacement the hash of the replacement
+    /// @param siblings of replacement in bottom-up order
+    function getRootAfterReplacementInDrive(
+        uint256 position,
+        uint256 log2SizeOfReplacement,
+        uint256 log2SizeOfDrive,
```
GCdePaula, line 36: What's the difference between these two `log2SizeOfReplacement` and `log2SizeOfReplacement`?

GCdePaula (commented):
I left a few questions!

cartesi-rollups/contracts/src/MerkleConstants.sol
guidanoli: Yes, the change in the machine state Merkle tree affects this.
This is reflected here, as $2^5=32$ bytes.

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: I'm assuming you meant `log2SizeOfReplacement` and `log2SizeOfDrive`.
The first is the size of the section of the drive that will be replaced, for example, some buffer in memory.
The second is the size of the whole drive, for example, the whole machine memory.

GCdePaula: Gotcha, so on most cases `log2SizeOfDrive` will be 64?

GCdePaula: I'm trying to understand how this function will be used/called.

cartesi-rollups/contracts/src/MerkleConstants.sol
guidanoli: I see, because the RISC-V word is still 8 bytes, right?
In this case, let's rename this to `LOG2_LEAF_SIZE`.

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: I think this function will not be used in `DaveConsensus`.

GCdePaula: Does it make sense to remove it?

GCdePaula:
Do you want to merge this before adding tests?

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: I wonder if it might be useful later. If not, we can remove it.
For now, I don't think there is a strong reason to remove it.

guidanoli:
> Do you want to merge this before adding tests?

Either way works for me. What do you prefer?

GCdePaula:
I think we can merge it and add tests later.

GCdePaula (approved):
LGTM!

PR #81, opened 2024-10-21
Refactor `IDataProvider`

### LINKED ISSUE #80: Refactor `IDataProvider`
As of now, the implementation of the `gio` function in the `DaveConsensus` contract uses a hard-coded tree height that covers 2 MB, but what we should do, instead is:

- Compute the Merkle tree of the smallest tree that fits the response data
- This tree must encompass at least 32 bytes
- The function should return, besides the Merkle root, the size of the response data (in bytes)

As it is already implemented, when computing the Merkle root of the response data, we should pad it with zeros to the next power of 2.

guidanoli:
Close #80

### DIFF
--- a/cartesi-rollups/contracts/src/Merkle.sol
+++ b/cartesi-rollups/contracts/src/Merkle.sol
@@ -59,6 +59,24 @@ library Merkle {
         return replacement;
     }
 
+    /// @notice Get the root of the smallest Merkle tree that covers a byte array.
+    /// @param data the byte array
+    /// @dev If data is smaller than the drive, it is padded with zeros.
+    /// @dev The smallest tree covers at least one leaf.
+    /// @dev See `MerkleConstants` for leaf size.
+    function getSmallestMerkleRootFromBytes(bytes calldata data) internal pure returns (bytes32) {
+        for (
+            uint256 log2SizeOfDrive = MerkleConstants.LOG2_LEAF_SIZE;
+            log2SizeOfDrive <= MerkleConstants.LOG2_MEMORY_SIZE;
+            ++log2SizeOfDrive
+        ) {
+            if (data.length <= (1 << log2SizeOfDrive)) {
+                return getMerkleRootFromBytes(data, log2SizeOfDrive);
+            }
+        }
+        revert("Data larger than memory");
+    }
+
     /// @notice Get the Merkle root of a byte array.
     /// @param data the byte array
     /// @param log2SizeOfDrive log2 of size of the drive
--- a/cartesi-rollups/contracts/src/DaveConsensus.sol
+++ b/cartesi-rollups/contracts/src/DaveConsensus.sol
@@ -41,9 +41,6 @@ contract DaveConsensus is IDataProvider {
     /// @notice GIO namespace for getting advance requests from the InputBox contract
     uint16 constant INPUT_BOX_NAMESPACE = 0;
 
-    /// @notice GIO response buffer size (2 MB)
-    uint256 constant LOG2_GIO_RESPONSE_BUFFER_SIZE = 21;
-
     /// @notice The input box contract
     IInputBox immutable _inputBox;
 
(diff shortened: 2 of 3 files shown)


cartesi-rollups/contracts/src/Merkle.sol
```diff
@@ -59,6 +59,24 @@ library Merkle {
...
+    function getSmallestMerkleRootFromBytes(bytes calldata data) internal pure returns (bytes32) {
+        for (
+            uint256 log2SizeOfDrive = MerkleConstants.LOG2_LEAF_SIZE;
+            log2SizeOfDrive <= MerkleConstants.LOG2_MEMORY_SIZE;
+            ++log2SizeOfDrive
+        ) {
+            if (data.length <= (1 << log2SizeOfDrive)) {
+                return getMerkleRootFromBytes(data, log2SizeOfDrive);
+            }
+        }
```
GCdePaula, line 76: There's a better way (which probably in practice shouldn't matter as much) which is `O(log(log(size)))` instead of `O(log(size))`.

This procedure is very similar to `clp2(x)`, which is defined as "the least power of 2 that is >= x". In other words, it is the `2^(ceil(log2(x)))`. The difference is that we want `log2_clp2(x)`, which is just `ceil(log2(x))`.

Here's a Lua code that uses `clz` (which I think we already have in Solidity) to compute `log2_clp2(x)`:

```lua
-- Returns number of leading zeroes of x. Shamelessly stolen from the book
-- Hacker's Delight.
local function clz(x)
  if x == 0 then return 64 end
  local n = 0
  if (x & 0xFFFFFFFF00000000) == 0 then
    n = n + 32; x = x << 32
  end
  if (x & 0xFFFF000000000000) == 0 then
    n = n + 16; x = x << 16
  end
  if (x & 0xFF00000000000000) == 0 then
    n = n + 8; x = x << 8
  end
  if (x & 0xF000000000000000) == 0 then
    n = n + 4; x = x << 4
  end
  if (x & 0xC000000000000000) == 0 then
    n = n + 2; x = x << 2
  end
  if (x & 0x8000000000000000) == 0 then n = n + 1 end
  return n
end

local function log2_clp2(x)
  assert(math.ult(0, x))
  return 64 - clz(x - 1)
end

for i = 1, (1 << 20) do
  local log2 = log2_clp2(i)
  assert(math.ceil(math.log(i, 2)) == log2)
  assert(i <= 1 << log2)
end
```

Note that it's common to define the `clp2(0)` as `0`. The above snippet on the other hand only accepts `x > 0`.

The _Hackers Delight_ is a great source for these bit shenanigans.

guidanoli: Thanks! Yeah, this is definitely more efficient in the worst case, but I'd like to benchmark the difference between the naïve approach (iterating, like I did) or this binary search approach. Maybe naïve is always better than binary, or binary is always better than naïve, or there is a point in which one is better than the other.

guidanoli: For the sake of not having the gas cost explode for larger inputs, I'd say we should go for the binary search option. :-)

cartesi-rollups/contracts/src/Math.sol
```diff
@@ -46,4 +46,59 @@ library Math {
 
         return n;
     }
+
+    /// @notice count leading zeros
+    /// @param x number you want the clz of
+    /// @dev this a binary search implementation
+    function clz(uint256 x) internal pure returns (uint256) {
```
guidanoli, line 53: This has been *shamelessly* copied from [`CartesiMath`](https://github.com/cartesi/solidity-util/blob/7fdd7d7547da79fb99ff8fb6b2a2843d91fb8276/packages/contracts/contracts/CartesiMathV2.sol#L113) with slight modifications.

- Renamed `_num` as `x`
- Changed visibility from `public` to `internal`

GCdePaula: We can implement `ctz` as a function of `clz` too:

```
256 - nlz(~x & (x - 1))
```

GCdePaula (dismissed):
LGTM! Left a comment, but it's fine as it is!

guidanoli:
Implemented `ctz` using `clz`, as you suggested, and added tests.

GCdePaula (approved):
LGTM!

PR #90, opened 2024-12-14
refactor: simplify `IDataProvider`

guidanoli:
- The `IDataProvider` implementation should be aware of epoch boundaries, and receive input indices relative to the epoch, to facilitate usage by tournament contracts
- `namespace` is not necessary because, as of now, it's just a constant
- `extra` is always `input` in the current stage
- Block numbers are not necessary, because input indices delimit epoch boundaries already
- Without thinking in terms of block numbers, we don't need to decode input blobs


GCdePaula (approved):
✨✨

PR #94, opened 2024-12-22
refactor: refactor build and tests pipeline

GCdePaula:
chore(prt-tests): remove `--platform` from Dockerfile


GCdePaula:
So far I've:
* introduced `just` and rewrote many `makefile`s into `justfile`s;
* reorganized directories (minor);
* built a development `Dockerfile` with nice caching and all dependencies (including the current branch of the Cartesi machine, which it builds), as well as building Dave.

One possible next step is removing the usage of docker in our make/just targets. Currently, we have a `Dockerfile` that installs all dependencies, builds the repo and run tests. Our `make` targets invokes docker (docker build then docker run).

I think a better workflow just invokes the intended commands, without docker, and assumes all dependencies are there (i.e. the target `just test` directly runs our test scripts; it assumes all the dependencies are there). Then we also provide a docker image which has these dependencies in case the developer doesn't have these dependencies installed in their system (`docker run dave:dev just test`).

GCdePaula:
* Implemented anvil deployments as a state dump and load, as explained [here](https://github.com/cartesi/dave/pull/93#discussion_r1892191971).
* Changed time from timestamps to block number, and fixed tests (`vm.warp` -> `vm.roll`). I think this is the most critical part of the PR. We'd planned on changing this; I didn't intend to change it now, but thew previous item required it.
* Rewrote the way we build Cartesi "programs". So far, only `echo` is using this new pipeline (check `test/programs`).
* Made node state directory configurable.
* Running integration tests outside docker on my machine:
```
just bind && just -f test/programs/justfile build-echo && just -f prt/tests/rollups/justfile test-echo
```

GCdePaula:
This PR is getting close to completion.

Feature-wise, the main change is from block timestamp -> block number, which is a delicate change. The rest is mainly refactoring and changes to how we invoke programs.

Running tests on the host: 
```bash
$ just build && just -f test/programs/justfile build-echo
$ just -f prt/tests/rollups/justfile test-echo
```

Running tests in docker:
```bash
$ just run-dockered just -f prt/tests/rollups/justfile test-echo
```

cartesi-rollups/node/machine-runner/src/lib.rs
```diff
@@ -16,14 +16,15 @@ use cartesi_machine::{break_reason, configuration::RuntimeConfig, htif, machine:
 use cartesi_prt_core::machine::constants::{LOG2_EMULATOR_SPAN, LOG2_INPUT_SPAN, LOG2_UARCH_SPAN};
 use rollups_state_manager::{InputId, StateManager};
 
-// gap of each leaf in the commitment tree, should use the same value as CanonicalConstants.sol:log2step(0)
+// gap of each leaf in the commitment tree, should use the same value as CanonicalConstants.sol:log2step(0)a
```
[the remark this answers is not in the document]
GCdePaula: Typo!

GCdePaula:
With newer foundry version, we started hitting smart contracts size limit. The solution (great catch @stephenctw!) is to add the option `--disable-code-size-limit` in anvil _both when storing and loading_.

An important consideration is whether this limit will appear when deploying to main net (or whether it's only our tests that hit this limit).

GCdePaula:
This ended up being a larger PR than I'd imagined. It's mostly refactoring, except from the change of timestamp -> blocknumber, which is quite major, but the smart contracts were well implemented enough that it was a simple change (but the clients required more effort).

cartesi-rollups/contracts/justfile
```diff
@@ -0,0 +1,50 @@
...
+        --sig 'run()' \
+        -vvvv 2>&1
+    forge script \
+        script/DaveConsensus.s.sol \
+        --fork-url {{ANVIL_ENDPOINT}} \
+        --code-size-limit 65536 \
+        --broadcast \
+        --sig 'run(bytes32,address)' \
+        {{INITIAL_HASH}} \
+        $(jq -r '.transactions[] | select(.transactionType=="CREATE").contractAddress' broadcast/InputBox.s.sol/{{ANVIL_CHAIN_ID}}/run-latest.json) \
```
guidanoli, line 49: To avoid this, we could join these two Solidity scripts into one.

GCdePaula: Good idea!

stephenctw: I think I purposely want them to be two separate blocks so that the the first epoch already have inputs to process

justfile
```diff
@@ -0,0 +1,58 @@
...
+bind-consensus:
+    just -f ./cartesi-rollups/contracts/justfile bind
+
+build-prt:
+    just -f ./prt/contracts/justfile build
+clean-prt-bindings:
+    just -f ./prt/contracts/justfile clean-bindings
+bind-prt:
+    just -f ./prt/contracts/justfile bind
+
```
stephenctw, line 28: Should we also add the `tests` related just targets? Something like 
```
test-prt-rollups-echo:
    just -f prt/tests/rollups/justfile test-echo
```
?

stephenctw:
> This ended up being a larger PR than I'd imagined. It's mostly refactoring, except from the change of timestamp -> blocknumber, which is quite major, but the smart contracts were well implemented enough that it was a simple change (but the clients required more effort).

I think there's one more place to update timestamp -> blocknumber
https://github.com/cartesi/dave/blob/feature/cleanup/prt/client-lua/player/reader.lua#L81

stephenctw (approved):

justfile
GCdePaula: I think we can add it in a later PR!

cartesi-rollups/contracts/justfile
GCdePaula: Ok! I've kept as @stephenctw prefer, and we can keep discussing this later. I'm fine either way.

PR #100, opened 2025-01-28
test: Merkle libs

guidanoli:


cartesi-rollups/contracts/src/Merkle.sol
GCdePaula, line 1: I'm wondering about this line on `getHashOfLeafAtIndex`:

```
keccak256(abi.encode(bytes32(data[start:])))
```

Is it taking a large slice, when it should be taking a 32-bytes slice (i.e. `data[start:end]`)?

GCdePaula (approved):
Great work :)

cartesi-rollups/contracts/src/Merkle.sol
guidanoli: Your concern is that it may be doing unnecessary `CALLDATALOAD` operations, right?
I've run this code through Forge's interactive debugger, and it only loads the first 32 bytes of the slice, and nothing more. Adding an extra `end` in the slice would only make this function use more gas only to achieve the same result.

PR #108, opened 2025-02-08
refactor: improve anvil startup and teardown

guidanoli:
Code inspired from [Bug Buster](https://github.com/crypto-bug-hunters/bug-buster/blob/9228d92d1d8a88d9d05c413fc051b436e617e82f/tests/bounties/src/adder/setup-exec-env.sh#L34).


guidanoli:
This has already been addressed by #164.

PR #112, opened 2025-02-24
Load contract addresses from deployment file

stephenctw:
Since we've done a couple big and small refactorings on the smart contracts. Their deployment order and addresses change from time to time. It becomes annoying to change the "default address" which we thought shouldn't change a lot. Hence I try to extract the smart contract addresses from `addresses` file that produced by our anvil script.

### DIFF
--- a/prt/tests/rollups/justfile
+++ b/prt/tests/rollups/justfile
@@ -1,10 +1,18 @@
+ECHO_DIR := "../../../test/programs/echo"
+
+INPUT_BOX := `head -n 1 ../../../test/programs/echo/addresses`
+CONSENSUS := `head -n 2 ../../../test/programs/echo/addresses | tail -n 1`
+
 # run PRT rollups test
 test MACH_PATH:
   rm -rf _state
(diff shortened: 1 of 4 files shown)


prt/tests/rollups/justfile
```diff
@@ -1,10 +1,18 @@
+ECHO_DIR := "../../../test/programs/echo"
+
+INPUT_BOX := `head -n 1 ../../../test/programs/echo/addresses`
```
stephenctw, line 3: I really want to replace `../../../test/programs/echo/addresses` here with `ECHO_DIR`. I've tried so many options but none of them work. Can use some help here!

GCdePaula (commented):
Looks good! I've left a comment and a question.

prt/tests/rollups/justfile
stephenctw: Yeah, it's working now. I just feel like we shouldn't have to repeat `../../../test/programs/echo` over and over. We can leave it as is and improve it when we figure out how to properly use the variable.

PR #118, opened 2025-03-14
Create `TransitionState.sol` contract 

stephenctw:
- Depend on this [PR](https://github.com/cartesi/machine-solidity-step/pull/76), will have to update submodule after it's merged.
- Also added `matchDeleted` event for explorer

### DIFF
--- a/prt/contracts/src/ITransitionPrimitives.sol
+++ b/prt/contracts/src/ITransitionPrimitives.sol
@@ -1,0 +1,32 @@
+// Copyright Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+pragma solidity ^0.8.0;
+
+import "step/src/AccessLogs.sol";
+
+/// @title ITransitionPrimitives
+/// @notice Contain three primitives that transitions machine state from s to s+1
+interface ITransitionPrimitives {
+    function step(AccessLogs.Context memory a)
+        external
+        pure
+        returns (AccessLogs.Context memory);
+
+    function reset(AccessLogs.Context memory a)
+        external
+        pure
+        returns (AccessLogs.Context memory);
+}
--- a/prt/contracts/src/ITransitionPrimitivesCmio.sol
+++ b/prt/contracts/src/ITransitionPrimitivesCmio.sol
@@ -1,0 +1,29 @@
+// Copyright Cartesi and individual authors (see AUTHORS)
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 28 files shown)


GCdePaula:
I have a design suggestion.

I think the interface `IStateTransition` should have a single function: `transitionState`, which receives the `IDataProvider`. I think this interface should be defined and implemented here. Then, the leaf tournament simply calls `transitionState`.

In fact, I think we should have two implementations of `IStateTransition`: one for rollup and another for compute (the compute implementation will ignore the `IDataProvider`).

This way, I think we don't need to change the solidity-step at all.

stephenctw:
Great suggestion indeed. I'll make the change.

GCdePaula (approved):
Looks great!

Are the contract sizes deployable with this? If so, it's ok to merge as is personally.

One suggestion, which (if we choose to implement) could be done in a later PR, is moving the `transition` function from the leaf tournament to the `IStateTransition`. In this idea, the step, reset and cmio wouldn't even be visible to the leaf tournament. We could even have two contracts that implement of `IStateTransition`: one for rollup and another for compute.

Nice work.

GCdePaula:
Ah sorry! Yeah, exactly like that ✨✨

Now, looking more carefully, I think there's an error in the STF (`transition`).

We can fix it now, or in a later PR, possibly TDD style (we write a test that catches it, then fix it).

The first issue is the revert when the data provider returns 0x0. I believe it shouldn't revert, only skip the cmio (but still step).

Second issue I'm less sure about, let's think more, but it's about the zero length check.

GCdePaula:
One advantage of this design is that we can now test transition in isolation.

prt/contracts/src/ITransitionPrimitives.sol
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+pragma solidity ^0.8.0;
+
+import "step/src/AccessLogs.sol";
+
+/// @title ITransitionPrimitives
+/// @notice Contain three primitives that transitions machine state from s to s+1
+interface ITransitionPrimitives {
```
GCdePaula, line 22: Naming suggestion: `IRiscVStateTransition`

prt/contracts/src/ITransitionPrimitivesCmio.sol
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+pragma solidity ^0.8.0;
+
+import "step/src/AccessLogs.sol";
+
+/// @title ITransitionPrimitives
+/// @notice Contain three primitives that transitions machine state from s to s+1
+interface ITransitionPrimitivesCmio {
```
GCdePaula, line 22: Naming suggestion: `ICMIOStateTransition`

prt/contracts/src/ITransitionState.sol
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+pragma solidity ^0.8.0;
+
+import "./IDataProvider.sol";
+
+/// @title ITransitionState
+/// @notice Transitions machine state from s to s+1
+interface ITransitionState {
```
GCdePaula, line 22: Naming suggestion: `IStateTransition`

prt/contracts/src/ITransitionState.sol
```diff
@@ -0,0 +1,29 @@
...
+    function transition(
```
GCdePaula, line 23: Naming suggestion: `transitionState`

stephenctw: Thank you naming master 😎

GCdePaula (approved):
✨✨✨

GCdePaula:
I've done some work on CI and testing on the machine bindings PR. Maybe some of the issues I encountered and fixed there are similar to the ones here.

GCdePaula (approved):

PR #102, opened 2025-01-29
feat(cartesi-machine): update machine bindings

GCdePaula:
New bindings for the new C API. Current commit is using release [v0.19.0-test2](https://github.com/cartesi/machine-emulator/releases/tag/v0.19.0-test2). I believe we should only merge this on a proper release. It's also possible to test without a release (i.e. latest main branch commit), but requires a local docker installation.

Remember to run `just clean-emulator`, since your local copy will probably have build artifacts of older versions. You need to have the local supporting files too: run`just setup`. To test, run `cargo test -p cartesi-machine`.

PR #121, opened 2025-03-19
Fix state transition function

GCdePaula:
I've changed some directories around, and the standard of includes to absolute paths.

(Tangent: I'm thinking on whether we should rename `ITournament` to something else, since `Tournament` does not implement `ITournament`. Maybe something like `ITask` and `ITaskFactory`?)

This is still a preliminary PR, I still need to review the offchain code analogous to the onchain STF. It's easier to see the changes related to the STF by ignoring the first commit of this branch.


GCdePaula:
This branch is based on `feature/update-bindings`. After we merge #102, I'll rebase on top of main, and change the target branch to main.

GCdePaula:
Ready to review!

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -95,7 +95,7 @@ contract DaveConsensusFactoryTest is Test {
             }
             if (entry.topics[0] == DaveConsensus.EpochSealed.selector) {
                 (
-                    uint256 epochNumber,
+                    ,
```
fabiooshiro, line 98: Not a block, of course. Why not _epochNumber?

GCdePaula: In solidity tuple deconstruction, for unused variables it's idiomatic to skip the name altogether instead of prepending an underscore.

test/programs/yield/.gitkeep
stephenctw, line 1: What is this directory for?

stephenctw (commented):
Amazing work! I left some comments, in general this looks really great!

stephenctw (approved):
🤘

PR #127, opened 2025-03-25
Add rollups echo test to CI

stephenctw:
@guidanoli Please let me know if you have any concern of adding `winnerCommitment` to the `canSettle` function return values.
CI is running three times as slow as my laptop. Anyways the test finishes in 15 minutes. I'm not sure if it should be enabled for all pushes, or only for tag/releases. What do you think @GCdePaula ?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,6 +82,7 @@ jobs:
           just test-rust-workspace
 
       - name: Test dave-rollups
+        if: startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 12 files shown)


GCdePaula:
Yeah, maybe only for tags. I want to try to optimize this later.

GCdePaula (approved):

stephenctw:
> Yeah, maybe only for tags. I want to try to optimize this later, then we can revisit this decision.

Just change it. Need reapproval, thanks!

PR #128, opened 2025-03-26
Add DaveConsensusFactory and tests

pedroargento:


prt/contracts/cannonfile.toml
```diff
@@ -0,0 +1,45 @@
+name = 'cartesi-prt-multilevel'
+version = '0.0.1'
+description = 'Cartesi PRT contracts'
+
+[deploy.RiscVStateTransition]
+artifact = "RiscVStateTransition"
+salt = "<%= zeroHash %>"
```
tuler, line 7: use `create2 = true` for all deployments

.gitmodules
```diff
@@ -15,7 +15,7 @@
 	url = https://github.com/cartesi/rollups-contracts
 [submodule "cartesi-rollups/contracts/lib/openzeppelin-contracts"]
 	path = cartesi-rollups/contracts/lib/openzeppelin-contracts
-	url = https://github.com/OpenZeppelin/openzeppelin-contracts
+	url = https://github.com/openzeppelin/openzeppelin-contracts
```
guidanoli, line 18: Despite GitHub organization names being case-insensitive, the canonical name of the OpenZeppelin org is `OpenZeppelin` (as it is now), so I don't see how this change would necessary.

cartesi-rollups/contracts/cannonfile.toml
```diff
@@ -0,0 +1,20 @@
+name = 'dave-consensus-factory'
```
guidanoli, line 1: Prefix this with `cartesi-`?

cartesi-rollups/contracts/package.json
guidanoli, line 1: - Missing `@usecannon/cli` as dev dependency?
- Many fields here are unnecessary
- Update license

cartesi-rollups/contracts/src/DaveConsensusFactory.sol
```diff
@@ -0,0 +1,66 @@
...
+    constructor(IInputBox _inputBox, ITournamentFactory _tournament) {
+        inputBox = _inputBox;
+        tournamentFactory = _tournament;
+    }
+
+    function newDaveConsensus(address appContract, Machine.Hash initialMachineStateHash)
+        external
+        returns (IDataProvider)
+    {
+        IDataProvider daveConsensus =
```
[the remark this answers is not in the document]
guidanoli: I agree. The `IDataProvider` interface shouldn't even be mentioned here or in the interface.

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
guidanoli, line 1: I don't see why we need to mock the input box and tournament factory contracts.
Wouldn't it be easier to just use the actual implementations?

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -0,0 +1,122 @@
...
+    MockTournamentFactory _tournamentFactory;
+    MockTournament _tournament;
+    address _appContract;
+    Machine.Hash _initialMachineStateHash;
+
+    function setUp() public {
+        _inputBox = new MockInputBox();
+        _tournamentFactory = new MockTournamentFactory();
+        _factory = new DaveConsensusFactory(_inputBox, _tournamentFactory);
+        _appContract = address(0x1234);
```
guidanoli, line 73: The `forge-std` lib has this function `vm.addr` which derives an address from a private key, which I normally use to generate seemingly random addresses.

```suggestion
        _appContract = vm.addr(1);
```

cartesi-rollups/contracts/src/DaveConsensusFactory.sol
```diff
@@ -0,0 +1,66 @@
...
+
+import {DaveConsensus} from "./DaveConsensus.sol";
+import {IDataProvider} from "prt-contracts/IDataProvider.sol";
+import {ITournamentFactory} from "prt-contracts/ITournamentFactory.sol";
+import {IInputBox} from "rollups-contracts/inputs/IInputBox.sol";
+import {Machine} from "prt-contracts/Machine.sol";
+
+/// @title Dave Consensus Factory
+/// @notice Allows anyone to reliably deploy a new `IDataProvider` contract.
+contract DaveConsensusFactory {
```
guidanoli, line 16: Maybe define an interface for this factory?

prt/contracts/cannonfile.toml
guidanoli, line 1: Inconsistent order of `create2` and `salt` fields.

prt/contracts/package.json
guidanoli, line 1: Same comments about the other `package.json` apply here.

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -115,6 +81,18 @@ contract DaveConsensusFactoryTest is Test {
                 address emittedAddress = abi.decode(entry.data, (address));
                 assertEq(emittedAddress, address(daveConsensus));
             }
+            if (entry.topics[0] == DaveConsensus.EpochSealed.selector) {
+                (
+                    uint256 epochNumber,
+                    uint256 inputIndexLowerBound,
+                    uint256 inputIndexUpperBound,
+                    Machine.Hash initialMachineStateHash,
```
guidanoli, line 89: You could also validate the initial machine state hash, since you're providing it as an argument to the instantiate function of the factory contract.

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -115,6 +81,18 @@ contract DaveConsensusFactoryTest is Test {
                 address emittedAddress = abi.decode(entry.data, (address));
                 assertEq(emittedAddress, address(daveConsensus));
             }
+            if (entry.topics[0] == DaveConsensus.EpochSealed.selector) {
+                (
+                    uint256 epochNumber,
```
guidanoli, line 86: You could validate the epoch number being zero.

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -115,6 +81,18 @@ contract DaveConsensusFactoryTest is Test {
...
+                    uint256 inputIndexLowerBound,
+                    uint256 inputIndexUpperBound,
```
guidanoli, line 88: You could validate the lower bound being zero,
and, before instantiating the contract, add `n` inputs (where `n` is a limited fuzzy parameter), and check that the upper bound is `n`.
Fuzzy parameters can be restricted to a range through the function [`bound`](https://book.getfoundry.sh/reference/forge-std/bound).

pedroargento: I agree about validating the initial machine state hash. The other validations I think it belongs more to DaveConsensus tests than to the factory (or some sort of integration test). What do you think?

guidanoli: If we view the factory as a black box, then how do I know that it is passing down the `IInputBox` parameter to the `DaveConsensus` instance. To add `n` inputs and check this input index upper bound is one way.

cartesi-rollups/contracts/test/DaveConsensusFactory.t.sol
```diff
@@ -40,23 +39,36 @@ contract DaveConsensusFactoryTest is Test {
         _initialMachineStateHash = Machine.Hash.wrap(0x0);
     }
 
-    function testNewDaveConsensus(address appContract, address _tournamentAddress) public {
+    function testNewDaveConsensus(address appContract, address _tournamentAddress, uint256 numberOfInputs) public {
+        numberOfInputs = bound(numberOfInputs, 0, 10);
         vm.recordLogs();
         _tournamentFactory.setAddress(_tournamentAddress);
+        for (uint256 i = 0; i < numberOfInputs; ++i) {
```
guidanoli, line 46: We don't need to initialize counters with `0` because that is their default value.

GCdePaula (approved):
🚀🚀🚀

PR #142, opened 2025-04-15
Re-enable compute and add to CI

stephenctw:
- implement `get_logs_compute` in both Rust and Lua
- make use of snapshot to speed up the tests (only for compute, doesn't affect rollups)
- adapt `just` to build compute programs and run compute tests
- add `test-simple` to CI

I believe adding compute test to CI can help us catch early bug in the future whenever we make changes to prt code.


prt/client-lua/utils/json.lua
GCdePaula, line 1: I think your formatter changed this file. I think we should leave it as is!

prt/client-rs/core/src/db/compute_state_access.rs
GCdePaula, line 1: We gotta change these "compute" names eventually!

GCdePaula (approved):
Great work!

prt/client-lua/utils/json.lua
stephenctw: Sorry I'll undo the change and force push back.

GCdePaula (approved):

PR #163, opened 2025-05-15
fix(docker): fix libslirp not found issue

GCdePaula:
I've also improved how we install just and foundry, to stop building from source because it's too slow.

### DIFF
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -1,25 +1,25 @@
 # syntax=docker.io/docker/dockerfile-upstream:1.12.1-labs
 
+#### prepare tools
+FROM rust:1.85 AS tools
+# foundry
+RUN set -eux; \
+  apt-get update && apt-get install -y --no-install-recommends curl ca-certificates git; \
+  curl -L https://foundry.paradigm.xyz | bash; \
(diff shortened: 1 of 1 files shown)


test/Dockerfile
```diff
@@ -1,37 +1,45 @@
 # syntax=docker.io/docker/dockerfile-upstream:1.12.1-labs
 
+#### prepare tools
+FROM rust:1.85 AS tools
+# foundry
+RUN set -eux; \
+  apt-get update && apt-get install -y --no-install-recommends curl ca-certificates git; \
+  curl -L https://foundry.paradigm.xyz | bash; \
+  ~/.foundry/bin/foundryup --install stable; \
+  install -Dm755 ~/.foundry/bin/* /usr/local/bin/
```
guidanoli, line 10: Applying the following diff got the Docker build to work for me.

```suggestion
ARG DEBIAN_FRONTEND=noninteractive
RUN <<EOF
set -eu
apt-get update && apt-get install -y --no-install-recommends curl ca-certificates git
curl -L https://foundry.paradigm.xyz | bash
~/.foundry/bin/foundryup --install stable
install -Dm755 ~/.foundry/bin/* /usr/local/bin/
EOF
```

PR #168, opened 2025-05-19
fix(node): fix sharing provider between async runtimes

GCdePaula:


GCdePaula:
Yeah! There's been an intermittent transport error `TransportError(…, hyper::Error(IncompleteMessage))`, which is making the node crash.

After investigating, the issue seems to be related to the latest tokio rework.

Currently, we instantiate one single provider in tokio runtime A. This runtime then ends, and we instantiate runtimes B and C.

Both B and C receive the provider created by now-dead A.

It seems request+hyper except that everything be in the same runtime. Or at least that the runtime that created everything doesn't die. Possibly both. The point is that sharing causes issues. Here's a [related issue](https://github.com/seanmonstar/reqwest/issues/2501) with a comment by the owner of request.

This patch instantiates one provider per service/worker, inside the worker's own tokio runtime.

stephenctw (approved):
Looks great and all tests passed on my side locally. I wonder how to verify the issue is being addressed. Just put it running on testnet?

GCdePaula:
Yeah, my plan was creating a new release and seeing if the test deployment stops crashing.

PR #175, opened 2025-05-30
Eliminate inner tournament

GCdePaula:


prt/contracts/src/tournament/abstracts/NonRootTournament.sol
```diff
@@ -47,6 +49,38 @@ abstract contract NonRootTournament is Tournament {
...
+        (bool _hasDanglingCommitment, Tree.Node _danglingCommitment) =
+            hasDanglingCommitment();
+
+        // If the tournament is finished but has no winners,
+        // inner tournament can be eliminated
+        if (!_hasDanglingCommitment) {
+            return true;
+        }
+
+        (Clock.State memory clock,) = this.getCommitment(_danglingCommitment);
```
guidanoli, line 69: If you change the visibility of the `getCommitment` from `external` to `public`, you can call the function directly as if it were an `internal` function. This reduces gas cost because Solidity then compiles this to a `JUMP` instead of a `CALL` to itself.

```suggestion
        (Clock.State memory clock,) = getCommitment(_danglingCommitment);
```

GCdePaula:
@stephenctw @guidanoli

I'm happy with the Solidity code so far. I think there's one edge case missing to test (fair clock deduction). As such, the solidity part is ready to review.

I'll start implementing the node code soon.

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -398,10 +404,29 @@ abstract contract Tournament {
...
 
+    /// @notice returns if and when tournament was finished.
+    /// @return (bool)
+    /// - if the tournament can be eliminated
+    /// - the time when the tournament was finished
+    function timeFinished() public view returns (bool, Time.Instant) {
+        if (!isFinished()) {
+            return (false, Time.ZERO_INSTANT);
+        }
+        /// - the
```
guidanoli, line 419: Is this comment complete?

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -55,6 +55,7 @@ abstract contract Tournament {
     //
     Tree.Node danglingCommitment;
     uint256 matchCount;
+    Time.Instant lastMatchElimination;
```
guidanoli, line 58: Since the internal function is called `deleteMatch` and the event is called `matchDeleted`, why not name this state variable `lastMatchDeletion` instead, just for naming consistency?

prt/contracts/src/tournament/libs/Clock.sol
```diff
@@ -122,11 +122,23 @@ library Clock {
...
     }
 
+    /// @notice Deduct duration from a clock and set it to paused.
+    /// The clock must have time left after deduction.
+    function deduct(State memory state, Time.Duration deduction)
+        internal
+        pure
+        returns (State memory)
+    {
+        assert(state.startInstant.isZero());
```
guidanoli, line 137: I wonder if we should use custom errors instead of assertions.
I know these conditions should never be true, but in case of bugs, it's good to have easy traceability of errors through selectors.

prt/contracts/test/Util.sol
guidanoli, line 1: Why were these variables made public?
I don't see their getter functions being called anywhere.

prt/contracts/src/tournament/abstracts/NonRootTournament.sol
```diff
@@ -18,34 +18,70 @@ abstract contract NonRootTournament is Tournament {
...
+
+    using Clock for Clock.State;
+
     /// @notice get the dangling commitment at current level and then retrieve the winner commitment
     /// @return (bool, Tree.Node, Tree.Node)
     /// - if the tournament is finished
     /// - the contested parent commitment
-    /// - the dangling commitment
-    // TODO: handle when no one wins, i.e no one joins the tournament
+    /// - the winning inner commitment
```
guidanoli, line 30: Mention the new return value in the function documentation?

prt/contracts/src/tournament/abstracts/NonRootTournament.sol
```diff
@@ -18,34 +18,70 @@ abstract contract NonRootTournament is Tournament {
...
+
+        // If the tournament is finished but has no winners,
+        // inner tournament can be eliminated
+        if (!_hasDanglingCommitment) {
+            return true;
         }
+
+        // We know that, after `winnerCouldHaveWon` plus  winner's clock.allowance has elapsed,
+        // it is safe to elminate the tournament.
+        (bool finished, Time.Instant winnerCouldHaveWon) = timeFinished();
```
guidanoli, line 81: Would it be possible, instead of calling `isFinished` and `timeFinished`, to just call `timeFinished`, and use the Boolean return value the same as the return value of `isFinished`, so you don't need the `assert` and you don't call (indirectly) `isFinished` twice?

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -398,10 +404,29 @@ abstract contract Tournament {
     }
 
     /// @return bool if the tournament is over
-    function isFinished() internal view returns (bool) {
+    function isFinished() public view returns (bool) {
         return isClosed() && matchCount == 0;
     }
 
+    /// @notice returns if and when tournament was finished.
+    /// @return (bool)
```
guidanoli, line 412: Missing `Time.Instant` return value in documentation.

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -88,6 +89,10 @@ abstract contract Tournament {
         _;
     }
 
+    constructor() {
+        lastMatchElimination = Time.currentTime();
+    }
```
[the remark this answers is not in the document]
GCdePaula: Nice catch. Yeah, let's use this `startInstant` approach.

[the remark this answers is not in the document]
GCdePaula: I don't think we actually need to do it, but it's more sound.

I don't think we need because either there were eliminated matches (in which case it will be overwritten), or not (in which case `max(lastMatchElimination, startInstant + allowance)` would already result in `startInstant + allowance`).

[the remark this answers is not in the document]
GCdePaula: Which would explain why this piece of code not doing what we want has no observable impact.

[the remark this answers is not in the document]
GCdePaula: Or maybe we should just leave it at `0` 🤔

What are your thoughts @guidanoli?

prt/contracts/src/tournament/abstracts/NonRootTournament.sol
GCdePaula: Nice, much better.

prt/contracts/src/tournament/abstracts/Tournament.sol
GCdePaula: Added.

prt/contracts/src/tournament/abstracts/NonRootTournament.sol
GCdePaula: Done.

prt/contracts/test/Util.sol
GCdePaula: Removed public

prt/contracts/src/tournament/libs/Clock.sol
guidanoli: I know that defining a custom error for each invariant is tiring.
We can evaluate this for a later iteration.
Let's leave it as is for now.

stephenctw (approved):
off-chain part looks great to me!

PR #180, opened 2025-06-04
chore(prt): update challenge period to 1 week

GCdePaula:


GCdePaula:
I can't make the tests work with a 50_000 block week 😔

We'll need to split the deployment constants, creating a new one specific for local testing.

GCdePaula:
> Any reason why those are hardcoded constants, and not configurable at contract construction? So we can use different values for different environments.

https://github.com/cartesi/dave/pull/182

@tuler

PR #182, opened 2025-06-05
Parameterize time constants

guidanoli:
This PR makes the following changes:

- Add `matchEffort` and `maxAllowance` as constructor arguments of `CanonicalTournamentParametersProvider`
- Makes the PRT contracts' Cannonfile pick values for these parameters based on the chain ID
- Adjusts tests (which instantiate `CanonicalTournamentParametersProvider`)
- Fix "Dave PRT for Cartesi Rollups" Cannonfile referencing wrong Cannon package versions


guidanoli:
You can test what it would be like to deploy to Ethereum Mainnet like so:

```sh
# On prt/contracts
pnpm cannon build --wipe --chain-id 1 --dry-run --impersonate-all
```

For other networks, you may choose a different value passed to the `--chain-id` option.
E.g.: 11155111 for Ethereum Sepolia, 13370 for Anvil (make sure `anvil` is running in the background)

For unsupported networks, the Cannonfile will fail.
That is because I've listed the chain type (mainnet, testnet, or devnet) for each chain explicitly, as well as the average block time (which varies, e.g., for OP Mainnet it is 2 seconds).

GCdePaula (approved):
🚀🚀🚀

PR #186, opened 2025-06-09
fix(node): make node clean up partial snapshot in case of failure

GCdePaula:


stephenctw (commented):
The fix seems pretty solid. I wonder if we could add some test cases for it?

stephenctw:
I also noticed the compute tests are not working. I don't think it's broken by this PR, but should we fix it or it's not our top priority for now?

GCdePaula:
> should we fix it or it's not our top priority for now?

I'm not sure, but I'd say it's not our priority now.

PR #191, opened 2025-07-03
Attach contract artifacts to GH releases

guidanoli:
This PR adds a CI job that uploads contract artifacts to GH releases.
It is necessary for the `rollups-node` to generate bindings for the Dave contracts.
We also generate an Anvil state dump with all Cartesi Rollups and PRT contracts.
Once this is merged, I will create a dummy release to test the new CI job and start adapting the `rollups-node` code on a separate, to-be-created `next/3.0` branch.
FYI @vfusco


GCdePaula (approved):
LGTM!

PR #193, opened 2025-07-10
Remove `compute` from `client-rs`

stephenctw:
- Completely remove `compute` code from Rust
- Remove nested `compute_tree` table from db
- Rename some files/variables/functions


prt/client-rs/core/src/db/dispute_state_access.rs
```diff
@@ -125,79 +120,82 @@ impl ComputeStateAccess {
...
-        let tx = conn.unchecked_transaction()?;
-        for digest_and_leaf in compute_trees {
-            compute_data::insert_compute_tree(
-                &tx,
-                digest_and_leaf.0.slice(),
-                digest_and_leaf.1.iter(),
-            )?;
-        }
-        tx.commit()?;
+        leafs: Vec<(Vec<u8>, u64)>,
```
GCdePaula, line 165: Can we type `(Vec<u8>, u64)` as `Leaf` instead?

GCdePaula (commented):
I'm thinking whether the `client-rs/executable` still makes sense.

---

This is the signature of `react_once`:
```
pub async fn react_once(
        &mut self,
        arena_sender: &impl ArenaSender,
    ) -> Result<PlayerTournamentResult>;
```

I'm thinking whether we should remove the `arena_sender` from the arguments. Maybe make putting it as a field of `Player`? If so, we'd need to add a generic.

Also, we are instantiating a new `Player` each react in the `epoch_manager`, right? I think we should instantiate it once, and call `react_once` at every iteration. Also, it might make sense to remove the current `react`, and rename `react_once` to `react`.

---

There is a `mut commitments: CommitmentMap` argument. Does it make sense to remove it? I think so; I'm not seeing its usefulness.

---

I think we maybe should refactor to remove `MachineCommitmentBuilder`. Here's a possible plan:

* add the initial machine path to the db;
* make `commitment::build_machine_commitment` only return the leaves;
* make the state access directly return the merkle tree instead of the leafs.

This way, the state access has a function that returns the full commitment. It should:
* try to get the leaves;
* if not present, compute them by calling `commitment::build_machine_commitment` and add the leaves to the db;
* Finally, call `commitment:: build_machine_commitment_from_leafs` with the leaves (either freshly computed or cached in the DB).

stephenctw:
>There is a mut commitments: CommitmentMap argument. Does it make sense to remove it? I think so; I'm not seeing its usefulness.
I see it's being used to retrieve parent's commitment and does something with that.

PR #197, opened 2025-08-27
Implement revert in `client-rs`

stephenctw:


prt/client-rs/core/src/machine/instance.rs
```diff
@@ -142,19 +149,19 @@ impl MachineInstance {
...
+        }
+        self.snapshot_path = snapshot_path;
+
         while self.input_count < input_count {
-            let input = db.input(self.input_count)?;
-            if input.is_none() {
-                self.input_count = input_count;
-                break;
-            }
+            // snapshot the machine state before feeding the input
```
GCdePaula, line 159: I think we need to do the snapshot, and also do a revert if needed somewhere in this loop.

prt/client-rs/core/src/machine/instance.rs
```diff
@@ -260,13 +287,25 @@ impl MachineInstance {
...
+                if self.revert_if_needed()? {
+                    let new_snapshot_path =
+                        db.work_path.join(format!("{}", self.root_hash()?.to_hex()));
+                    if !new_snapshot_path.exists() {
+                        self.machine.store(&new_snapshot_path)?;
+                    }
+                    if self.snapshot_path.exists() {
+                        std::fs::remove_dir_all(&self.snapshot_path)?;
+                    }
+                    self.snapshot_path = new_snapshot_path;
```
GCdePaula, line 306: The body of this `if` could perhaps be inside `revert_if_needed`.

I think this loop should be something like:

```
self.machine.run(target_physical_cycle)?;
// here, there are four possible reasons that `run` returned

// reason one: halted
if self.is_halted()? {
    panic!("run break with halt");
}

// reason two: manual yield
if self.is_yielded()? {
    self.revert_if_needed()?;
    break;
}

// reason three: reached target
if self.physical_cycle()? == target_physical_cycle {
    break;
}

// reason four: automatic yield
// continue loop
```

prt/client-rs/core/src/machine/instance.rs
GCdePaula: I think this is being done on `run`.

PR #204, opened 2025-10-20
Enrich events args for UI

stephenctw:
- Fix all events naming style (use Pascal)
- add extra event args
- index more event args
- name all event args
- update client-lua and client-rs accordingly

### DIFF
--- a/prt/contracts/src/tournament/abstracts/Tournament.sol
+++ b/prt/contracts/src/tournament/abstracts/Tournament.sol
@@ -92,12 +92,18 @@ abstract contract Tournament {
     // matches existing in current tournament
     mapping(Match.IdHash => Match.State) matches;
 
-    enum MatchDeletedReason {
-        TIMEOUT_ONE,
-        TIMEOUT_TWO,
-        BOTH_ELIMINATED,
-        SUBGAME_WINNER
+    enum MatchDeletionReason {
+        STEP,
+        TIMEOUT,
+        CHILD_TOURNAMENT
     }
+
+    enum WinnerCommitment {
+        NONE,
+        ONE,
+        TWO
+    }
+
     //
     // Events
     //
@@ -110,8 +116,10 @@ abstract contract Tournament {
     );
     event MatchDeleted(
         Match.IdHash indexed matchIdHash,
-        MatchDeletedReason reason,
-        Tree.Node indexed winnerCommitment
+        Tree.Node indexed one,
+        Tree.Node indexed two,
+        MatchDeletionReason reason,
+        WinnerCommitment winnerCommitment
     );
     event CommitmentJoined(
         Tree.Node commitment,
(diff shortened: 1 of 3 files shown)


ZzzzHui:
> Hey @ZzzzHui , I tried running the integration test with bond feature, yet got this error:
> 
> ```
> lua: ./test_env.lua:147: player fail to resume with error: ../../client-lua/player/sender.lua:101: Failed to get bond value: Error: server returned an error response: error code 3: execution reverted, data: "0x"
> ```
> 
> Could you help me with this?

Hey Stephen. Of course. Let me take a look

stephenctw:
> Hey @ZzzzHui , I tried running the integration test with bond feature, yet got this error:
> 
> ```
> lua: ./test_env.lua:147: player fail to resume with error: ../../client-lua/player/sender.lua:101: Failed to get bond value: Error: server returned an error response: error code 3: execution reverted, data: "0x"
> ```
> 
> Could you help me with this?

Seems like a foundry bug with 1.1.0 version.

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -72,14 +72,32 @@ abstract contract Tournament {
     // matches existing in current tournament
     mapping(Match.IdHash => Match.State) matches;
 
+    enum MatchDeletedReason {
+        TIMEOUT_ONE,
+        TIMEOUT_TWO,
+        BOTH_ELIMINATED,
+        SUBGAME_WINNER
+    }
```
guidanoli: In the case of `BOTH_ELIMINATED`, `winnerCommitment` has no meaning.

Another option would be to split `SUBGAME_WINNER` into `SUBGAME_WINNER_ONE` and `SUBGAME_WINNER_TWO`, just like it is currently being done with timeout. This way, the `winnerCommitment` event parameter would not be necessary.

Also, to make naming more consistent, should we rename `SUBGAME` as `CHILD_TOURNAMENT`?

Also, should we draw a distinction between step and child tournament win? If so, there should be reasons `STEP_WINNER_ONE` and `STEP_WINNER_TWO`.

stephenctw:
> Will approve once conflicts are resolved!

Did rebase and resolved conflicts.

prt/contracts/test/Tournament.t.sol
```diff
@@ -31,8 +31,11 @@ contract TournamentTest is Util {
     address player0 = vm.addr(1);
     address player1 = vm.addr(2);
 
-    event matchCreated(
-        Tree.Node indexed one, Tree.Node indexed two, Tree.Node leftOfTwo
+    event MatchCreated(
```
guidanoli: Instead of re-declaring the event, you can access the event definition from the original contract or library and emit it like so:

```solidity
emit ContractOrLibrary.MatchCreated(...)
```

GCdePaula (approved):
LGTM!

PR #210, opened 2025-10-30
fix WAL

ZzzzHui:


prt/client-rs/core/src/db/dispute_state_access.rs
```diff
@@ -89,6 +95,11 @@ impl DisputeStateAccess {
                     .map_err(anyhow::Error::from)
                     .unwrap();
 
+                connection
+                    .query_row("PRAGMA journal_mode=WAL;", [], |_| Ok(()))
+                    .map_err(anyhow::Error::from)
+                    .unwrap();
+
```
GCdePaula: WAL is persistent per db:
https://sqlite.org/wal.html#persistence_of_wal_mode

I think as such we only need to set it during creation, which is this code I've highlighted. The other path, which loads an existing database, can likely be deleted. I'm not sure though. What do you think? Can you verify?

PR #223, opened 2025-11-13
Small improvements

vfusco:


guidanoli:
The build is failing because the Cannon package was published to the Cannon registry...
When we make the change on Cartesi Rollups to not use Cannon for devnet build, we will do the same here on the Dave repo and then we can merge your PR. What do you think?

PR #224, opened 2025-12-03
Build Honeypot snapshot with devnet addresses

guidanoli:
This PR changes the Honeypot snapshot build process.
Instead of downloading the snapshot from GitHub, it:

- Downloads the source code from GitHub
- Creates a configuration file with the devnet addresses
- Compiles the Honeypot source code with such config
- Copies the snapshot created from source to the expected directory

It also makes some small adjustments elsewhere.

Following this PR, there needs to be further changes:

- Enrich the test suite with non-reverting inputs (deposits and withdrawals)
- Fix Docker environment build


ZzzzHui:
should probably just use the tag v3.0.0 instead of commit hash, in order to fix the CI

PR #225, opened 2025-12-08
refactor(prt-contracts): merge (non)rootTournament into Tournament

stephenctw:
Now we have a much simpler abstract structur - only (non)LeafTournament left. There's no more dimond inheritance and circular depedency. I think this is a great start, and we can discuss whether we should go further on this simplification process.


prt/client-lua/player/reader.lua
```diff
@@ -360,13 +360,17 @@ function Reader:read_match(address, match_id_hash)
...
+    local sig = "innerTournamentWinner()(bool,bytes32,bytes32,(uint64,uint64))"
     local ret = self:_call(address, sig, {})
+    assert(#ret >= 3, "innerTournamentWinner returned insufficient values")
 
     local winner = {
         has_winner = helper.str_to_bool(ret[1]),
-        commitment = Hash:from_digest_hex(ret[2]),
-        dangling = Hash:from_digest_hex(ret[3]),
+        parent_commitment = Hash:from_digest_hex(ret[2]),
+        dangling_commitment = Hash:from_digest_hex(ret[3]),
```
guidanoli: I see `tournament_winner.commitment` being accessed [here](https://github.com/cartesi/dave/blob/1f1249a9d9cadae5ed3b250485ff0c9b6a31f2c6/prt/client-lua/player/strategy.lua#L303), where `tournament_winner` is either the return value of `inner_tournament_winner` or `root_tournament_winner` (see [here](https://github.com/cartesi/dave/blob/1f1249a9d9cadae5ed3b250485ff0c9b6a31f2c6/prt/client-lua/player/state.lua#L62-L66)). In order to make `inner_tournament_winner` more similar to `root_tournament_winner`, shouldn't we rename `dangling_commitment` as just `commitment`? (We can leave the `parent_commitment` entry.)

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -105,15 +105,81 @@ abstract contract Tournament is ITournament {
...
+    function _nonRootTournamentArgs()
         internal
         view
         virtual
-        returns (bool, Machine.Hash, Machine.Hash);
+        returns (ITournament.NonRootArguments memory)
+    {
+        // Default implementation for root tournaments (level == 0)
+        // Non-root tournaments (level > 0) must override this
+        return ITournament.NonRootArguments({
```
guidanoli: The definition of the `NonRootArguments` struct is directly accessible in the context of the `Tournament` contract because it inherits from the `ITournament` interface. There is no need to access it through `ITournament`.

```suggestion
        returns (NonRootArguments memory)
    {
        // Default implementation for root tournaments (level == 0)
        // Non-root tournaments (level > 0) must override this
        return NonRootArguments({
```

PR #227, opened 2025-12-09
fix: improve leaf index validation

henriquemarlon:
With the stronger fuzzer introduced in Foundry v1.5.0, a new failing fuzz case was detected in:

```Merkle.getHashOfLeafAtIndex(bytes calldata data, uint256 leafIndex)```

The issue is caused by this line:

```uint256 start = leafIndex << MerkleConstants.LOG2_LEAF_SIZE;```

For very large leafIndex values, this shift can silently wrap around modulo `2^256`, making start small again. This may cause the condition start < data.length to incorrectly pass, leading to an invalid hash being returned instead of the pristine leaf.

Although this does not affect deployed contracts (where leafIndex is safely bounded), it breaks fuzz tests on Foundry v1.5.0.

This PR adds a reversible shift check to prevent wrap-around:

`if (start < data.length && (start >> MerkleConstants.LOG2_LEAF_SIZE) == leafIndex)`

This guarantees that the shift did not overflow and restores correct behavior under fuzzing.

### DIFF
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -8,7 +8,7 @@ RUN <<EOF
 set -eu
 apt-get update && apt-get install -y --no-install-recommends curl ca-certificates git
 curl -L https://foundry.paradigm.xyz | bash
-~/.foundry/bin/foundryup --install v1.4.0
+~/.foundry/bin/foundryup --install v1.5.0
 install -Dm755 ~/.foundry/bin/* /usr/local/bin/
 EOF
 # cargo binstall (to install just)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/Dockerfile [hunk 1]
```diff
@@ -8,7 +8,7 @@ RUN <<EOF
 set -eu
 apt-get update && apt-get install -y --no-install-recommends curl ca-certificates git
 curl -L https://foundry.paradigm.xyz | bash
-~/.foundry/bin/foundryup --install stable
+~/.foundry/bin/foundryup --install v1.5.0
```
guidanoli: The CI is currently using Foundry 1.4.3.
If you want to bump Foundry to 1.5.0, then please also update this line:

https://github.com/cartesi/dave/blob/1f1249a9d9cadae5ed3b250485ff0c9b6a31f2c6/.github/actions/setup-tools/action.yml#L15

You should then run `forge fmt` on the Forge projects.

PR #228, opened 2025-12-10
Adjustments to tournament simplification PR

guidanoli:
Hey @stephenctw, I've collected some suggested changes into a PR for your appreciation.
Let's discuss here the changes, so we can converge. :-)

### DIFF
--- a/prt/contracts/src/ITournament.sol
+++ b/prt/contracts/src/ITournament.sol
@@ -431,6 +431,12 @@ interface ITournament {
         view
         returns (TournamentArguments memory);
 
+    /// @notice Returns non-root tournament arguments
+    function nonRootTournamentArgs()
+        external
+        view
+        returns (NonRootArguments memory);
+
     /// @notice Check whether a match can be won by timeout.
     /// @param matchId The match ID
     function canWinMatchByTimeout(Match.Id calldata matchId)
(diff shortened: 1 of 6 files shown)


prt/contracts/src/ITournament.sol
```diff
@@ -431,6 +431,12 @@ interface ITournament {
         view
         returns (TournamentArguments memory);
 
+    /// @notice Returns non-root tournament arguments
+    function nonRootTournamentArgs()
```
guidanoli: I have added this function to the `ITournament` interface, because the `NonRootArguments` struct was not being used there. Also, it might be useful to expose this struct to the node.

prt/contracts/src/tournament/abstracts/Tournament.sol
```diff
@@ -170,16 +170,6 @@ abstract contract Tournament is ITournament {
 
     function _totalGasEstimate() internal view virtual returns (uint256);
 
-    //
-    // Helper Functions
-    //
-
-    /// @notice Check if this is a root tournament (level == 0)
-    function _isRootTournament() internal view returns (bool) {
```
guidanoli: I have removed this function because it was not being used.

stephenctw (approved):

PR #229, opened 2025-12-15
Improve Honeypot PRT test framework/suite

guidanoli:
This PR introduces the following changes:
- Apply @vfusco changes from #223
- Test deposit and withdrawal inputs against Honeypot PRT
- Dump post-test Anvil state for every test program/script combination

Now, when you run any test, it produces a `anvil_<PROGRAM>_<SCRIPT>.json` file.
For example, running `just test honeypot simple` generates `anvil_honeypot_simple.json`.


ZzzzHui (commented):
looking great, Gui. Just a few comments :)

PR #233, opened 2026-01-12
improve bond refund

ZzzzHui:


ZzzzHui:
Thanks Stephen!!

GCdePaula:
Nice!

Can you explain why the change in the smart contracts broke things, and why your fix fixed them?

stephenctw:
> Nice!
> 
> Can you explain why the change in the smart contracts broke things, and why your fix fixed them?

I might be wrong, but seems Alloy's estimate gas function was too conservative, hence many transactions were reverted silently because of running out of gas(after the bond improvements). You can also tell that the lua-client doesn't need to change a line. Meaning that "cast send" estimate tx gas pretty accurate.

PR #235, opened 2026-01-30
fix(docker): improve dockerized environment setup

henriquemarlon:


guidanoli:
Closing this, as #254 has already fixed the Dockerfile.
Also, #269 removed the dependency from pnpm.

PR #239, opened 2026-02-10
fix: update bindings, fix tests

stephenctw:
@GCdePaula Could you help taking a second look at the Rust-bindings. They are mostly done by AI agent, which works now, but you might want to make some changes.

### DIFF
--- a/prt/tests/rollups/setup_path.lua
+++ b/prt/tests/rollups/setup_path.lua
@@ -1,7 +1,19 @@
--- setup client-lua path
-package.path = package.path .. ";../common/?.lua"
-package.path = package.path .. ";../../client-lua/?.lua"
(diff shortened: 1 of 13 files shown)


prt/tests/rollups/setup_path.lua
GCdePaula: Do we need these changes?

GCdePaula (commented):
Why the "serde default" everywhere? Like, what changed in the new machine version that required us to add this "serde default"? Can you also explain what this default does?

stephenctw:
> Why the "serde default" everywhere? Like, what changed in the new machine version that required us to add this "serde default"? Can you also explain what this default does?

Apparently the machine now makes a lot of fields/configs optional. There are scenarios that the returned config is not "complete", and the jaon parser will crash of missing fields. So adding this default avoids parser error

PR #240, opened 2026-02-11
fix: check SHA-256 hash of release artifact

guidanoli:
It's good practice to check the hash of a downloaded file, to avoid supply-chain attacks.
Besides this change, this PR also:

- Forces `wget` to write the file to `add-generated-files.diff` by passing the `-O` option. Without such option, `wget` may avoid writing over `add-generated-files.diff` and instead write to `add-generated-files.diff.SUFFIX` where `SUFFIX` is some suffix added by `wget` to avoid collision with an already-existing file.

- Removes the `add-generated-files.diff` file after applying it. This ensures the Git submodule at `machine/emulator` stays clean after the repository is setup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,8 +138,10 @@ jobs:
         working-directory: ./machine/emulator
         run: |
           make bundle-boost
-          wget https://github.com/cartesi/machine-emulator/releases/download/v0.19.0/add-generated-files.diff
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
GCdePaula: Should you instead run the justfile target you updated in this CI step?

GCdePaula (approved):
Left a minor comment!

.github/workflows/build.yml
guidanoli: The CI workflow already did `wget`, so I didn't want to change the workflow.
If we can use the just target, we can do that in a separate PR.

PR #259, opened 2026-04-06
feat: avoid reentrancy attacks on PRT

### LINKED ISSUE #255: Check reentrancy in `joinTournament`
As an excess of caution, we should perhaps check for reentrancy in `joinTournament`:

```
require(!locked, ReentrancyDetected());
```

guidanoli:
Closes #255

### DIFF
--- a/prt/contracts/src/tournament/Tournament.sol
+++ b/prt/contracts/src/tournament/Tournament.sol
@@ -107,10 +107,20 @@ contract Tournament is ITournament {
         _;
     }
 
+    /// @notice Acquires the lock at the start,
+    /// and releases the lock at the end.
+    /// Avoids reentrancy attacks.
+    modifier withLock() {
+        _acquireLock();
+        _;
+        _releaseLock();
+    }
+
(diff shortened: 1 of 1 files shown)


stephenctw (approved):
Awesome work!

prt/contracts/src/tournament/Tournament.sol
GCdePaula, line 391: This is a state change after an external call. In this case, it's harmless, but perhaps worth it to document.

GCdePaula (approved):
Nice!

I think all the locks (including the existing ones) are actually not necessary, but a good idea regardless. It's defensive programming. It could be interesting to document that.

Approved!

prt/contracts/src/tournament/Tournament.sol
guidanoli: Yes, this is the only part of the call that make it not compliant to the Checks-Effects-Interactions pattern. That's why it's being protected by a lock.

guidanoli: Added documentation in bec647b.

PR #260, opened 2026-04-06
Custom errors and Rollups contracts bump

### LINKED ISSUE #257: Convert error strings to custom errors
It is easier for the node to handle custom errors, compared to standard error strings.
Below are the error strings that I could find in the source code and where they are located.

- `prt`
  - `contracts`
    - `src`
      - `tournament`
        - `libs`
          - `Clock.sol`
            - a paused clock can't timeout
            - can't advance clock with no time left
            - can't create clock with zero time
            - clock is not initialized
            - clock is initialized
          - `Match.sol`
            - match does not exist
            - match is not finished
            - match is not ready to be finalized
            - match can't be advanced
            - matches are not equal
            - match doesn't exist
      - `types`
        - `Tree.sol`
          - `child nodes don't match parent`
          - `tree node doesn't exist`
- `cartesi-rollups`
  - `contracts`
    - `src`
      - `Merkle.sol`
        - Replacement smaller than leaf
        - Drive larger than memory
        - Replacement larger than drive
        - Position is not aligned
        - Proof length does not match
        - Drive smaller than leaf
        - Drive larger than memory
        - Data larger than drive
        - stack error
      - `PristineMerkleTree.sol`
        - Height out of bounds
        - stack error

The custom errors under `prt/contracts` can be defined under `ITournament`, where we have already moved errors from the `Match` library before. As for the errors under `cartesi-rollups/contracts`, we can define them in an independent interface (e.g. `IMerkleErrors`) and make `IDaveConsensus` inherit from it.

guidanoli:
This PR makes several changes:

- Adds documentation to all existing custom errors in `ITournament`
- Renames the parameters of some errors in `ITournament`
- De-duplicates some errors in `ITournament`
- Converts error strings into custom errors (defined in `ITournament`)
- Calls `canBeEliminate` internally (and not through `this.`)
- Bumps Rollups contracts from 2.2.0 to 3.0.0-alpha.3

This last change is quite dense, so I'll break it down even further:

- Skip deployment of test tokens (as they are now present in Rollups SDK devnet)
- Remove local source code of test tokens
- Remove `CannonDeployments.sol` (as `ApplicationFactory` is now part of the build)
- Add `WithdrawalConfig` param to Dave app deployment entrypoints
- Make `DaveConsensus` implement `getLastFinalizedMachineMerkleRoot`
- Make `DaveConsensus.settle` revert if app is foreclosed
- Use Keccak256, Math, and Merkle libraries from Rollups SDK
- Remove local source code and tests of Merkle and Math libraries
- Make `IDaveConsensus` inherit from `BinaryMerkleTreeErrors` (defines errors that may be raised by `LibBinaryMerkleTree` from the Rollups SDK contracts)
- Improves the tests of `DaveAppFactory` (more similar to Rollups SDK contract tests)
- Adjusts Rust and Lua test codes (because of new deployment signatures, and inputs being added after deployment)

Closes #257


stephenctw (approved):
Looking great!

prt/contracts/src/ITournament.sol
```diff
@@ -188,44 +188,209 @@ interface ITournament {
...
+    /// function applied over an agreed-upon state has produced a
+    /// post-state that differs from that of the match commitment.
+    /// @param whichCommitment Which of the two commitments is wrong
+    /// @param computedPostState The post-state computed by the state-transition function
+    /// @param commitedPostState The post-state contained within the commitment
     error WrongFinalState(
-        uint256 commitment, Machine.Hash computed, Machine.Hash claimed
+        uint256 whichCommitment,
+        Machine.Hash computedPostState,
+        Machine.Hash commitedPostState
```
GCdePaula: `commitedPostState` -> `committedPostState`

GCdePaula (approved):
LGTM! I found a minor typo.

prt/contracts/src/ITournament.sol
guidanoli: Nice catch! Fixed on 7d5aeac.

PR #261, opened 2026-04-10
feat: update `REVERT_ROOT_HASH_ADDRESS`

stephenctw:
Resolving checkpoint address change mentioned [here](https://github.com/cartesi/machine-solidity-step/pull/85#issuecomment-4215887223).

### DIFF
--- a/cartesi-rollups/node/state-manager/src/rollups_machine.rs
+++ b/cartesi-rollups/node/state-manager/src/rollups_machine.rs
@@ -10,7 +10,10 @@ use cartesi_prt_core::machine::constants::{
 use crate::{CommitmentLeaf, Proof};
 use cartesi_machine::{
     config::runtime::{HTIFRuntimeConfig, RuntimeConfig},
-    constants::{break_reason, pma::TX_START},
+    constants::{
+        break_reason,
+        pma::{SHADOW_REVERT_ROOT_HASH_START, TX_START},
+    },
     error::{MachineError, MachineResult},
     machine::Machine,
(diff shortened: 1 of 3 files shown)


cartesi-rollups/node/state-manager/src/rollups_machine.rs
```diff
@@ -45,7 +45,7 @@ pub const STRIDE_COUNT_IN_EPOCH: u64 = 1
     << (LOG2_INPUT_SPAN_TO_EPOCH + LOG2_BARCH_SPAN_TO_INPUT + LOG2_UARCH_SPAN_TO_BARCH
         - LOG2_STRIDE);
 
-pub const CHECKPOINT_ADDRESS: u64 = 0x7ffff000;
+pub const CHECKPOINT_ADDRESS: u64 = 0xfe0;
```
edubart: Instead of having this hardcoded, you could use `CM_AR_SHADOW_REVERT_ROOT_HASH_START` from the C API starting with emulator `v0.20.0`. Also we are calling this `REVERT_ROOT_HASH` on the emulator.

edubart (changes requested):

edubart:
This is probably blocked by the PR bumping the emulator https://github.com/cartesi/dave/pull/241

mpolitzer:
I updated the emulator bump PR ~~but there is still one broken test, will investigate tomorrow.~~ all tests are green.

ref. https://github.com/cartesi/dave/pull/241

PR #265, opened 2026-05-06
fix: bump Debian and Boost on CI and Dockerfile

guidanoli:
In #241, we have bumped the Machine Emulator from 0.19.0 to 0.20.0, but we forgot to bump the dependencies necessary to build it in the CI and in the Dockerfile. This led the CI to fail building the node binaries for the `v3.0.0-alpha.1` release. We should patch this so that we at least have binaries available on GitHub Releases.

- Debian 12 (Bookworm) -> 13 (Trixie)
  - GCC 12 -> 14 (fully supports C++23)
  - Lua 5.4.4 -> 5.4.7

- Boost 1.81 -> 1.83

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,6 @@ jobs:
           - arch: arm64
             target: aarch64-unknown-linux-gnu
             os: ubuntu-24.04-arm
-    if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ${{ matrix.os }}
     container:
       image: rust:1.90-trixie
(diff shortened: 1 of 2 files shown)


endersonmaia (approved):
LGTM

But maybe we could let the build-release-node job to run for PRs, and just the last two steps on the jobs would be guarded by the tag.

This way we don't have to wait for the tag to see the error.

GCdePaula (approved):

guidanoli:
Appealing to @endersonmaia's suggestion, I've pushed a commit that makes CI always build the node. :-)

PR #270, opened 2026-07-08
Claim staging

### LINKED ISSUE #137: Improve `IOutputsMerkleRootValidator` tests in `DaveConsensus`
* Add test cases where Merkle proof fails for each possible reason;
* Test exactly the TX buffer leaf value, instead of using an address space with repeated states;
* Create function that generates siblings from random seed---instead of fuzzing of the output tree (and repeating it to generate an address space), let the siblings be fuzzed;
* Check that outputs Merkle root is invalid before settlement and valid after settlement.

guidanoli:
## Summary

This PR introduces claim staging to `DaveConsensus`: instead of settling an epoch in a single transaction, the tournament result is first *staged* and only *accepted* (settling the epoch) after either a configurable claim staging period has elapsed or all designated **sentries** have confirmed the result. This adds a safety window in which independent validators can cross-check a tournament's outcome before it becomes final, while still allowing fast settlement when all sentries agree. To ensure early epoch settlement, a designated sentry manager can rotate sentries if any of their credentials are compromised.

Closes #137

### Contracts
- Split `settle`/`canSettle` into staging and acceptance steps
- Anyone can stage a finished tournament's result; emits `EpochStaged`
- Staged results are accepted after the claim staging period elapses
- Acceptance is immediate if all sentries agree with the staged machine state hash
- Sentries are appointed upon deployment and assigned sequential IDs from 1 to N
- Sentries can be rotated by the sentry manager via `rotateSentry`; emits `SentryRotation`
- Sentries vouch for a post-epoch machine state hash via `submitSentryClaim`; emits `SentryClaim`
- New getters for the claim staging period, sentry manager and sentry-related state
- New errors for staging, acceptance, sentry rotations, and sentry claim requirements
- Added staging info to `getCurrentSealedEpoch` view function return values
- `IDaveAppFactory` functions take `claimStagingPeriod`, `sentryManager`, and `sentries`
- Tests extended to cover staging, sentry claims, sentry rotations, and acceptance
- Index `epochNumber` parameter in `EpochSealed` event
### Node
- `epoch-manager` updated to follow the new two-step flow with sentry claims
### Tests
- `blockchain-reader` test utils and the PRT rollups Lua test framework updated
### Chores
- Bump Foundry from 1.4.3 to 1.5.1
- Bump Honeypot from 3.0.0 to 3.0.1

### DIFF
--- a/.github/actions/setup-tools/action.yml
+++ b/.github/actions/setup-tools/action.yml
@@ -12,7 +12,7 @@ inputs:
   foundry-version:
     description: 'Foundry version to install'
     required: false
-    default: 'v1.4.3'
+    default: 'v1.5.1'
   setup-qemu:
     description: 'Whether to setup QEMU for riscv support'
     required: false
--- a/cartesi-rollups/contracts/script/Deployment.s.sol
+++ b/cartesi-rollups/contracts/script/Deployment.s.sol
@@ -1,7 +1,7 @@
 // (c) Cartesi and individual authors (see AUTHORS)
 // SPDX-License-Identifier: Apache-2.0
 
-pragma solidity ^0.8.8;
+pragma solidity ^0.8.30;
 
 import {BaseDeploymentScript} from "prt-contracts/../script/BaseDeploymentScript.sol";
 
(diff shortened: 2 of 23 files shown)


stephenctw (approved):
LGTM! Solid work!

guidanoli:
Merging this as both Stephen and Coutinho (off-line) approved.

PR #272, opened 2026-07-20
Proto Sling node

GCdePaula:
## What this is

A ground-up rewrite of the rollups node, landed as five reviewable commits on top of `next/3.0`. The headline: the node becomes a single crate where **one spec-oracled engine** serves both the machine runner's forward schedule and the dispute path's random access — previously two hand-maintained implementations of the same consensus-critical leaf semantics, kept in agreement only by e2e luck. It also ports the node to the staged settlement protocol (stage / sentry claim / accept) that landed on this base branch.

- `feat(node)!` — the node: one crate, one engine, the boundary store, the staged settlement port. Absorbs `prt/client-rs` and `common-rs` (deleted; the node was their only consumer).
- `test(e2e)` — the Lua harness: 23 scenarios (kill/respawn points, concurrent multi-sybil matches, on-chain timeout deletion), a parallel battery runner, hardened primitives. The Lua client stays an independent commitment oracle.
- `build` ×2 — the `just` gate (`just check`, `just doctor`), devnet fingerprinting, CI triggers + emulator build caching, and `just logged` (true-exit-code runner).
- `docs` — the knowledge base (`docs/`) and the campaign ledger (`docs/plans/`): design decisions with reasons, incident reports, and an internal audit ledger.

## Where review attention pays most

In order of consensus-criticality:

1. **`src/engine/`** — the meta-cycle geometry and leaf conventions (ustep spans, checkpoints, idle churn, revert). This is THE trust core. Read `docs/computation-hash.md` first; the conventions are pinned by 15 spec tests against a brute-force oracle, and the revert semantics were verified against `AdvanceStatus.sol` directly (the previous node and the Lua oracle shared a wrong assumption for months — contracts are the only source of truth).
2. **`src/epoch_manager/`** — the staged settlement port: three guarded steps per tick; sentries claim the locally computed post-epoch hash, never the staged value. The winner asserts are deliberate panics ("notify all users").
3. **`src/storage/`** — the trust boundaries in the schema itself: four mutation classes enforced by SQL triggers, write-once collision tripwires that panic (never retry), the frontier's strict prefix-bounding, and the boundary store's stage+rename crash discipline (`docs/plans/snapshots.md`).
4. **`src/tournament/`** — reader/sender/witness encodings vs the contracts. `docs/plans/node-audit.md` records an internal adversarial audit: 10 findings with dispositions and 56 explicit clean checks (every sender argument order and witness byte layout traced against contract consumption order).

The e2e harness (`prt/tests/rollups/`) deserves a pass on its trust story: the oracle is self-anchored (own machine lineage, chain inputs only) and judges the node strictly from outside.

## How it's verified

`just check` (fmt, luacheck, clippy `-D warnings`, unit + integration tests) green; the full 23-scenario battery green multiple times, including after the rebase onto this base (perf at parity with pre-rebase baselines). Layered nets: the spec oracle on enumerable structures, golden fixtures, committed chain recordings decoded through production bindings, and the independent Lua commitment oracle on every e2e epoch. Consensus-critical migrations were characterization-first: the old implementation's outputs were pinned as fixtures before each switch, and the differentials retired with their subjects.

## Breaking / operational notes

- Node state dirs from earlier prototypes are incompatible: the schema refuses them loudly (wipe and rebuild). No deployed-network impact; the node is pre-release.
- `prt/client-rs` and `common-rs` are gone (absorbed).
- CI now triggers on `pull_request` plus pushes to `main`/`next/*`/tags — feature branches get CI through PRs (forks finally get CI at all). First run populates the emulator cache; the speedup shows from the second run.

## Deliberately not here

The halt/exception contracts rework and its node-side dependents (the runner's halted-input wedge is a deliberate stopgap until that semantics exists on-chain); the settle-revert taxonomy and the rest of audit round 2; the per-scenario CI matrix. All tracked with their reasons in `docs/plans/`.


stephenctw (approved):
LGTM!

PR #273, opened 2026-07-20
Harden PRT dispute-game contracts

GCdePaula:
## Summary

This PR contains the smart-contract side of the PRT review. The corresponding node-reader adaptation is isolated in `feature/node-reader-update` and will follow separately, keeping this review focused on contract behavior and evidence.

### Fixes

- Terminal bond recovery now pays at most one bond and is idempotent. A failed recipient payment preserves the claimer and full balance so settlement remains permissionlessly retryable. This fixes an exploit where an attacker could brick the tournament.
- Refund and payment callbacks can no longer block tournament progress: calls have bounded gas, copy no return data, skip zero-value transfers, and preserve state where retry is required. This fixes an exploit where a claim made by a malicious smart contract could brick the tournament.
- After a successful recovery, any balance remaining after the one-bond payment is burned. Pooled reserves behind eliminated commitments therefore cannot be recycled into another Sybil campaign, preserving a real marginal cost for each attack. This fixes a bug that made recurrent Sybil attacks cheap, where losing Sybil bonds could be recycled into later disputes.
- A sealed-leaf timeout could restore time already spent by the survivor, while proof and timeout paths could overlap. Settlement now uses live clock balances and one four-way classifier shared by timeout views, mutations, and leaf-proof eligibility. This fixes a bug that degraded tournament liveness.
- Refund allocations and bonds previously rested on incomplete gas measurements. The new harness covers storage-sensitive branches, both orientations, and full-stack leaf proofs, then rounds allocations conservatively. Bonds derive from the maximum legal work reserve, so gas-table changes propagate into the required deposit. This fixes a bug where bond value was underestimated.

### Security and correctness testing

Testing is the largest part of this PR, accounting for more than 60% of its changed lines. The campaign replaces geometry-coupled coverage and implementation-shaped helpers with injectable fixtures, independent models, focused regressions, stateful positive and rejection campaigns, recursive lifecycle and accounting scenarios, adversarial callback recipients, and full-stack proof witnesses.

The resulting suite passes 235 PRT dispute tests across 52 suites, 4 downstream Rollups contract tests, 3 deterministic STF tests, both STF fuzz properties at 256 cases each, and 30 release-pinned gas witnesses. Scoped coverage is 98.43% lines, 98.47% statements, 49.31% branches, and 100% functions.

### Simplifications

- The front-loaded clock design is gone, substituted by a back-loading clock design. Pairing previously granted `matchEffort` to both participants before either responded, introducing bankable, capped balance changes. A successful response now receives a non-bankable discount after the fact: clock balances never increase, unused grants cannot be banked, timeout calibration is more direct and easier to reason about (major win), the tournament's worst-case delay bound is tighter, and pairing performs less storage work.
- `Clock` owns single-clock arithmetic and distinguishes checked from saturating subtraction; `MatchClocks` owns pair-level transitions and timeout policy.
- `Match` now expresses its phases directly, branches once during bisection and sealing, and stores sealed final states in commitment order.
- Match IDs now only locate mapping slots; existence is checked from the stored `isInit` flag. The old nonzero-hash check accepted arbitrary IDs, allowing nonexistent or deleted matches to produce plausible view results. Views and phase-specific transitions now reject uninitialized slots consistently.
- Internal library interfaces now receive the dimensions they operate on explicitly instead of depending on canonical configuration objects. Tests can now inject small or adversarial geometries while production geometry remains separately pinned.

## Review map

The PR changes 102 files and +25,495/-1,864 lines, but tests and documentation account for 93.1% of the changed lines.

| Area | Files | Changed lines | Share |
| --- | ---: | ---: | ---: |
| Tests and fixtures | 50 | 16,836 | 61.54% |
| Documentation and guidance | 30 | 8,634 | 31.56% |
| Production Solidity | 13 | 1,539 | 5.63% |
| Tooling and witness support | 9 | 350 | 1.28% |

Frozen review and calibration records contribute 6,107 added lines. The seven core logic files are 1,245 changed lines, or 4.55% of the PR.

A focused review can start with `Time.sol`, `Clock.sol`, and `MatchClocks.sol`; continue through `Match.sol` and `Tournament.sol`; and finish with `Bond.sol`, `Gas.sol`, and the corresponding focused and recursive suites.

---

<details>
<summary> Reproduce </summary>

## Reproduce

Run from the repository root:

```sh
just doctor
just check
just test-smart-contracts
just prt-contracts::test-stf
just prt-contracts::test-stf-fuzzy
just test-prt-gas
just prt-contracts::coverage
just prt-contracts::compatibility-hashes
```

The authoritative gas calibration requires a clean worktree and Forge 1.5.1:

```sh
just measure-prt-gas
```

Refund allocations are bounded subsidies, not receipt-exact reimbursement or safety bounds. The final post-refactor measurement required no further production-constant changes: `ADVANCE_MATCH` matches its rounded recommendation, inner sealing retains 1,000 gas of headroom, and `WIN_LEAF_MATCH` is calibrated to the largest retained canonical InputBox witness.
</details>

---

This PR intentionally changes the PRT ABI; backward compatibility is not required for this release line. External function selectors and semantic storage layout remain stable, but `TournamentArguments` changes shape and terminology, `PartialBondRefund` changes its event signature, timeout errors are renamed, impossible-state errors are removed, and `DaveConsensus.supportsInterface` changes behavior. The clone immutable-argument encoding and production bytecode change, so generated bindings, deployment artifacts, and CREATE2-derived addresses must be regenerated and reviewed before release. Node/Lua adaptation remains in the follow-up PR.

---

<details>
<summary>Compatibility details</summary>

## Compatibility details

`getMatch` has one intentional semantic change: after sealing, `leftNode` always contains commitment one's final state and `rightNode` always contains commitment two's final state. The previous height-parity-dependent interpretation no longer applies.

Raw readers must determine the Match phase before interpreting the overloaded tuple. Their corresponding update is handled in `feature/node-reader-update`.

</details>


stephenctw (approved):
LGTM!

Leaving the node/Lua timeout alignment and getMatch sealed-tuple reader update for the follow-up branch (`feature/node-reader-update` / timeout plan).

cartesi-rollups/contracts/test/DaveAppFactory.t.sol
guidanoli, line 1: (Refactor Suggestion) [DaveAppFactory.t.sol.patch](https://github.com/user-attachments/files/30514808/DaveAppFactory.t.sol.patch)

prt/contracts/src/tournament/libs/Time.sol
```diff
@@ -13,6 +16,8 @@ library Time {
     Instant constant ZERO_INSTANT = Instant.wrap(0);
     Duration constant ZERO_DURATION = Duration.wrap(0);
 
+    /// @notice Return the current EVM block-number coordinate.
+    /// @dev Deployment policy converts wall-clock durations into this coordinate.
```
guidanoli: (Documentation nit)

The words "policy" and "coordinate" sound a bit out of place.

```suggestion
    /// @notice Return the current EVM block number.
    /// @dev Deployment scripts converts wall-clock durations into EVM block numbers.
```

prt/contracts/src/tournament/libs/Clock.sol
```diff
@@ -15,167 +18,198 @@ library Clock {
...
-        } else {
-            // otherwise the allowance must be greater than the timespan from current time to start instant
-            return state.allowance
-                .gt(Time.timeSpan(Time.currentTime(), state.startInstant));
-        }
+    /// @notice Require an initialized paused clock.
+    /// @dev A violation after initialization is an internal phase-machine bug.
+    function requirePaused(State memory state) internal pure {
+        state.requireInitialized();
+        assert(!state.isRunning());
```
guidanoli, line 54: (Debugging nit)

Custom errors are easier to debug.

prt/contracts/src/ITournament.sol
```diff
@@ -35,16 +35,18 @@ interface ITournament {
     /// @param levels The number of tournament levels
     /// @param startInstant The start instant of the tournament
     /// @param allowance The time during which the tournament is open
-    /// @param maxAllowance The maximum time of a player clock
-    /// @param matchEffort The worst-case time to compute a commitment
+    /// @param maxAllowance The configured root allowance; parent-linked child
+    /// clocks inherit no more than this value
+    /// @param matchEffort The maximum elapsed-time discount earned by each
```
guidanoli: (Contract API Semantical Change)

Should we rename `matchEffort` as `responseBudget`?

prt/contracts/src/tournament/libs/MatchClocks.sol
```diff
@@ -0,0 +1,171 @@
...
+        Time.Duration responseBudget,
+        Time.Instant current
+    ) internal returns (Time.Duration) {
+        _pauseResponderAt(one, two, responseBudget, current);
+        return one.pausedAllowance().max(two.pausedAllowance());
+    }
+
+    /// @notice Pause the running responder, discounting its response.
+    /// @dev Every legal exit from active bisection discounts the responder
+    /// exactly once; the public verbs differ only in which clocks run next.
```
guidanoli: (Documentation nit)

The words "exit" and "verbs" sound a bit out-of-place here.
Wouldn't "action(s)" be a more fitting word in this context?

prt/contracts/src/tournament/libs/Bond.sol
```diff
@@ -0,0 +1,61 @@
...
+    /// @notice Largest configured terminal allocation shared by every level.
+    function terminalAllocation() internal pure returns (uint256) {
+        uint256 timeoutTerminal =
+            _max(Gas.WIN_MATCH_BY_TIMEOUT, Gas.ELIMINATE_MATCH_BY_TIMEOUT);
+        uint256 leafTerminal =
+            Gas.SEAL_LEAF_MATCH + _max(Gas.WIN_LEAF_MATCH, timeoutTerminal);
+        uint256 innerTerminal = Gas.SEAL_INNER_MATCH_AND_CREATE_INNER_TOURNAMENT
+            + _max(Gas.WIN_INNER_TOURNAMENT, Gas.ELIMINATE_INNER_TOURNAMENT);
+
+        return _max(timeoutTerminal, _max(leafTerminal, innerTerminal));
```
guidanoli: (Old behavior; non-blocking)

Post-calibration, `leafTerminal` evaluates to 6.3x the value of `innerTerminal`, and, because of `_max`, is used for all tournament levels. As a result, the bond value of all non-leaf tournaments is increased by 0.1851 ETH without need. If we add an `isLeafTournament` boolean parameter, we can easily lower this value. The call site on `Tournament.bondValue` can also easily provide this value through the already-existing `_isLeafTournament` internal function.

```suggestion
        return _max(timeoutTerminal, isLeafTournament ? leafTerminal :  innerTerminal);
```

prt/contracts/src/ITournament.sol
```diff
@@ -161,15 +151,17 @@ interface ITournament {
...
-    /// @param value The amount that was refunded
-    /// @param success Whether the refund was successful
-    /// @param ret The return data of the refund call
-    /// @dev In the case of a failed refund (success = false),
-    /// the argument `ret` may encode an smart contract error.
-    /// A refund should only fail if the recipient account
-    /// has any code (which can be an EOA, see EIP-7702).
+    /// @param value The computed refund, whether or not payment succeeded
+    /// @param success Whether the requested payment succeeded or was zero
+    /// @param ret Retained for event ABI compatibility; always empty
```
guidanoli: (Cleanup) Remove param?
The node doesn't seem to read this event.

guidanoli (commented):
Great PR overall! Just minor nits and refactor suggestions.

cartesi-rollups/contracts/test/DaveAppFactory.t.sol
GCdePaula: Thanks for the patch! The consolidation looked good, and we incorporated most of it:

- the settlement callback receiver is shared by the test fixture;
- the existing lifecycle test now covers recovery through a contract recipient;
- the bespoke `DaveConsensus` deployment and settlement application stub were removed;
- the staging gas ceiling is checked on the consolidated path; and
- the named regression is now a small deterministic wrapper around that path.

We made one deliberate change to the proposed ordering. When the recipient rejects payment, we leave settlement pending through `acceptStagedTournamentResult`, verify that consensus advances to a distinct next tournament, and only then enable payments and recover the retired tournament explicitly.

We also keep separate references to the old and new tournaments instead of overwriting the old one. This preserves both properties the regression is intended to pin: failed settlement cannot block epoch acceptance, and the retired tournament remains recoverable after the epoch advances.

prt/contracts/src/tournament/libs/Clock.sol
GCdePaula: Thanks! We’re keeping the `assert` intentionally.

This check represents an internal invariant: every legal contract transition must call this operation only while the clock is paused. A caller following the public state machine cannot legitimately trigger the failure. If it ever occurs in production, it means our own contract implementation violated its state-transition invariants.

This follows the project-wide distinction we use for failure handling:

- custom errors describe reachable, caller-facing failures that users and integrations may reasonably encounter and handle;
- assertions describe states that should be unreachable unless the contracts themselves contain a bug.

Modelling this as a custom error would make an impossible internal state look like a supported runtime failure and, under our interface conventions, expose an error that callers should never need to handle. A panic is therefore the intended signal here: the invariant was broken, rather than an ordinary invalid action being attempted.

The heuristic is: removing these asserts should not change the behavior of the code (except in case of bugs in the smart contracts themselves).

GCdePaula:
Thanks for the thorough review, Gui. The follow-up commits address the feedback and include a few related cleanups uncovered while auditing the same boundaries.

The API cleanup commit:

- renames `matchEffort` to `responseBudget`;
- removes the duplicate `TournamentArguments.maxAllowance`;
- removes the unused `PartialBondRefund.ret` parameter;
- gives the two timeout failures action-specific names;
- advertises `IDaveConsensus` through ERC-165;
- removes custom errors for impossible internal states and represents those states with assertions instead.

The heuristic for asserts is: removing these asserts should not change the behavior of the code (except in case of bugs in the smart contracts themselves).

On that last point, caller-controlled invalid inputs still fail through descriptive errors before reaching an assertion. We added regression tests for those ordering boundaries, including an unfinished registered child and a child entrant with an uncontested final state. `TournamentFailedNoWinner` remains a custom error because it represents a possible protocol outcome when the single-honest-player assumption does not hold.

The bond commit makes terminal work reserves role-specific. Leaf tournaments reserve for leaf terminal paths, while non-leaf tournaments reserve for inner-tournament paths. This reduces non-leaf bonds without changing any action refund cap or weakening the reserve argument. We also refreshed the calibration record and compatibility witnesses.

Finally, we incorporated the suggested factory-test refactor. The gas-exhausting settlement-recipient scenario now runs through the complete factory lifecycle, while preserving a deterministic pending-recovery regression and explicitly verifying that epoch acceptance advances before the retired tournament’s bond is recovered.

A final naming-only commit retains every invariant check while renaming the invariant-only helpers to `assertInitialized`, `assertPaused`, `assertRunning`, and `_assertBisection`. `requireUninitialized` remains a `require` because repeated initialization is caller-reachable.

The API changes above intentionally break ABI; backward compatibility is not required on this release line, and the corresponding node adaptation remains in the follow-up PR.

stephenctw (approved):
LGTM !

guidanoli (approved):
LGTM. 🚀

PR #274, opened 2026-08-05
Make tournament events authoritative, and teach the node to follow them

GCdePaula:
This PR redraws the contract-client boundary around one rule: **each fact should have one production authority**.

The tournament event stream now owns enumerable history and recursive structure: commitments, matches, child tournaments, and contract-authored match-elimination schedules. Narrow point views remain only where events are insufficient: tournament standing, the validator's one active match path, and terminal bond recovery. The Rust node builds those sources into one recursive `Dispute` instead of maintaining parallel event and view projections of the same state.

The result removes the old `O(matches)` point-read scan. Cleanup of clock-bearing matches is derived directly from events; tournament cleanup needs at most one standing read per relevant tournament; and match-detail reads follow only the Hero's recursive path.

This PR also retains two earlier pieces of the same contract-client campaign:

- bond recovery is an explicit permissionless action with a capability view, a winner bounty, and forfeited funds burned; and
- the rejected-input revert proof now reports the state the machine actually applies, fixing a bug that could make the honest node forfeit every `stf_revert`-shaped dispute.

## Why this shape

The earlier observer made both events and views describe candidate placement, live matches, clocks, phases, and recursive topology, then reconciled them on every tick. That was useful while the event model was incomplete, but it was not a good permanent abstraction: both answers came from the same contract and provider, so their agreement did not prove log completeness or provider honesty. It did add linear RPC work and a second authority for every fact.

The completed design keeps the useful boundaries while deleting that redundancy:

- **Events own history and structure.** Solidity mappings do not enumerate tournaments or matches, so creation, advancement, sealing, deletion, and child-creation events form the recursive dispute.
- **Contracts author cleanup deadlines.** `MatchCreated`, `MatchAdvanced`, and `LeafMatchSealed` carry the inclusive block at which that match becomes eliminable. A later response replaces the prior schedule; deletion cancels it.
- **Views answer only irreducible current questions.** `tournamentStanding()` handles tournament closure and carryover; match views are read only along the validator's engaged path; `bondRecovery()` classifies terminal recovery.
- **Mutators remain the final legality check.** Observation may become stale or another validator may win a race. The contract still revalidates every submitted action.

We deliberately keep the dispute in memory. The node persists the finalized raw event prefix and its cursor, reconstructs on restart, and retains one active recursive value. Moving this projection into SQLite, adding persistent collections, or creating aggregate contract views would add machinery before measurements justify it.

## Finalized Solid and Latest Foam

The reader maintains one finalized **Solid** dispute. Each tick it advances that prefix, commits the recognized logs and cursor, deep-clones the result, and recursively extends the clone through a sampled Latest height to obtain disposable **Foam**.

Foam is never promoted, reverse-applied, compared with the previous tail, or persisted. The next tick derives it again from Solid. A reorg or mixed latest response may suppress or delay one eager action, but it cannot mutate the finalized prefix.

The action policy follows the consequence of being wrong:

- joins derive their exact tournament and commitment from Solid; Latest may only suppress work that is already mined or no longer available;
- sentry votes and settlement content derive from finalized state;
- deadline-sensitive dispute responses and permissionless cleanup use Foam;
- recovery walks finalized tournament trees at a low cadence and may use Latest only to suppress an already-mined recovery, never to retire an epoch.

Every mutation shares one exclusive-signer transaction lane. A Hero action outranks cleanup; a tick emits at most one of them. Recovery runs only when no dispute or settlement mutation is ready, so maintenance cannot occupy a nonce needed by defense. Pending recoveries are rediscovered across epoch rotation and restart without a durable queue.

## Contract and compatibility changes

This is a coordinated, version-paired contract/client change:

- the observer surface is part of `ITournament`;
- ten legacy raw views are removed in favor of typed semantic results;
- `innerResult()` replaces the former result pair;
- `CommitmentJoined.commitment` is indexed;
- match lifecycle events now carry `eliminableAt`;
- leaf sealing has its missing structural event, `LeafMatchSealed`, and an event counter;
- `bondRecovery()` exposes the same terminal classification used by `tryRecoveringBond()`.

Storage layout and raw Match and Clock encodings are internal to a deployment generation: these contracts are not upgradeable, and no supported client reads raw slots. `leafMatchSealedCount` therefore lives with the other event counters, while the Solidity inspector and the one E2E clock probe follow the resulting layout. The storage fingerprint records implementation impact; it is not a promise to preserve slot numbers.

This is an explicit fresh-deployment cutover. The immutable clone payload changes from `levels: uint64` to `kind: TournamentKind`; the `MatchCreated` and `MatchAdvanced` signatures change, indexing `CommitmentJoined.commitment` changes its topics/data layout, and `LeafMatchSealed` is new. A fresh Tournament implementation, `MultiLevelTournamentFactory`, and dependent Dave bundle must deploy with matching bindings, Rust and Lua clients, and artifacts. Old and new implementations, factories, clones, event layouts, and persisted logs must not be mixed, and no live dispute crosses the cutover. There is no dual decoder.

As a final repository-layout cleanup, the two papers now live together under `docs/papers/`, while the repository-wide E2E harness lives under `test/e2e/`; package-local tests remain with their components. Git records both PDFs and all 41 harness files as renames, preserving their history.

## Diff distribution (approximate)

Counts are grouped by subsystem from `next/3.0...HEAD`; renames and cross-cutting files make the categories approximate.

| Category | Files | Insertions | Deletions | Review note |
|---|---:|---:|---:|---|
| Solidity implementation | 9 | +733 | -360 | **Small, security-critical contract and consensus surface** |
| Solidity tests | 28 | +1,977 | -136 | Observer, accounting, lifecycle, properties, and gas |
| Rust implementation | 42 | +11,901 | -3,185 | Recursive dispute, reader, observer, Hero, lane, recovery, and STF fix |
| Rust tests and fixtures | 9 | +135 | -24,129 | Large deletion is the retired chain recordings and second-fold oracle |
| Lua implementation | 16 | +4,764 | -731 | Coordinated semantic client and event-schedule support |
| Lua tests | 12 | +4,444 | 0 | Domain, fold, planner, adapter, and actor coverage |
| E2E harness and scenarios | 43 | +1,018 | -208 | History-preserving relocation plus GC, timeout, recovery, and sybil scenarios |
| Docs and guidance | 48 | +2,254 | -3,923 | Living architecture, active-plan pruning, review evidence, and agent routing |
| Tooling, build, and artifacts | 20 | +982 | -109 | CI, fingerprints, devnet generation, and gates |
| **Total** | **227** | **+28,208** | **-32,781** | |

## Review guide

The Solidity implementation is the protocol trust boundary: result-selection defects can become consensus failures, while clock or accounting defects can break liveness and resource bounds. The Rust node is a separate validator-safety boundary: a bad event transition, deadline decision, commitment, or proof can forfeit a dispute the honest validator should win.

### Solidity first

Read `docs/dispute-game.md`, `docs/prt-refund-accounting.md`, and `prt/contracts/AGENTS.md` beside `ITournament.sol`, `Tournament.sol`, and `MatchClocks.sol`.

<details>
<summary>Questions to verify</summary>

- Does every emitted `eliminableAt` equal the mutator's timeout boundary, inclusively, after creation, advancement, and leaf sealing?
- Can any response leave a stale schedule authoritative, or can deletion fail to cancel one?
- Does `LeafMatchSealed` complete the event history, with every white-box slot probe following the intentionally internal layout?
- Does `tryRecoveringBond()` preserve its revert, no-op, retry, payment, and burn behavior arm-for-arm with `bondRecovery()`?
- Do refunds, winner payment, and burn still conserve the tournament balance, including rounding and failed recipient callbacks?
- Are the revised gas allocations sufficient for the additional event and counter write?

</details>

### Recursive node model

Read `docs/plans/recursive-dispute-reader.md` and `docs/node-architecture.md`, then review `tournament/dispute.rs`, `reader.rs`, and `observer.rs`.

<details>
<summary>Questions to verify</summary>

- Is each event transition strict, block-atomic, and sufficient to reconstruct candidate placement, live matches, and child ownership?
- Can a child address enter the tree except through `NewInnerTournament` from an already trusted parent?
- Does recursive discovery handle a child created and used in the same block?
- Can a failed or inconsistent Latest load mutate Solid?
- Are resolved children retained for recovery but excluded from live Hero and cleanup traversal?
- Does match GC require no point reads, tournament GC remain bounded by tournament count, and Hero hydration read only one recursive path?
- Does the narrow observer convert ABI values into checked domain values without reintroducing whole-tree reconciliation?

</details>

### Action policy, recovery, and STF

Review the Hero actor/planners, epoch manager, recovery planner, transaction lane, and the revert path in `engine/`.

<details>
<summary>Questions to verify</summary>

- Does an exact join intent agree between Solid and Foam before submission?
- Can cleanup or recovery ever sit ahead of clock-bearing Hero work?
- Can Latest suppress recovery without permanently retiring an epoch?
- Do old pending recoveries survive epoch rotation and restart?
- Does every proving operation apply the same state transition as its non-proving twin, particularly the rejected-input revert?

</details>

## Validation

On the final local candidate, the following pass:

- `just check`
- `just prt-contracts::test-all`: 272 dispute tests, 3 STF FFI tests, and 2 fuzzy STF tests with 256 runs each
- `just test-prt-gas`: 18 PRT gas tests and 12 leaf-proof FFI gas tests
- `just rollups-contracts::test`: 5 tests
- contract coverage: 245 tests, 98.59% line coverage, and 100% function coverage
- `just test-lua-client`: 62 tests
- `just lint-lua`
- `just rollups-tests::test-sealed-leaf-timeouts`
- 13 focused recursive-reader regressions and all-target Rust Clippy with warnings denied
- `just test-rollups-echo`
- full `just test-rollups-honeypot`
- `just prt-contracts::compatibility-hashes`
- `just rollups-contracts::build-devnet`
- `git diff --check`

The compatibility report confirms the intended wire ABI, records the storage and deployment-bytecode impact, and places the Tournament runtime below the EIP-170 limit. The regenerated devnet bundle fingerprints the new deployment generation.

## Out of scope

- persisting the materialized dispute or Foam in SQLite;
- a Safe confirmation tier, latest rollback detector, or Foam promotion;
- aggregate or Multicall-shaped protocol APIs before RPC measurements justify them;
- tournament-level elimination schedules and fully event-derived bond recovery;
- decoding event topics from the previous Tournament deployment;
- changing tournament level coordinates;
- the emulator/solidity-step upgrade and block-builder submission backend.


prt/contracts/src/tournament/Tournament.sol
```diff
@@ -95,6 +95,9 @@ contract Tournament is ITournament {
     /// @dev Used by nested (non-leaf) tournaments
     mapping(ITournament => Match.Id) matchIdFromInnerTournaments;
 
+    // Appended after every existing persistent slot for layout compatibility.
+    uint256 leafMatchSealedCount;
```
guidanoli: This storage variable can live next to the other counters.
It's better for code organization and readibility.
There is no need to maintain storage layout compatibility.
Clients are not meant to read storage slots directly.
And we also don't do smart contract upgrades.
The public API for storage variables are the getter functions.

prt/contracts/src/tournament/Tournament.sol
```diff
@@ -417,18 +418,77 @@ contract Tournament is ITournament {
...
+    /// @notice The one recovery classification: the capability view and the
+    /// terminal mutation read the same arms in the same order.
+    /// @dev `claimer` and `payment` are populated only for `RECOVERABLE`.
+    /// The payment is one bond plus a tenth of the forfeited residual as a
+    /// defender's bounty, capped by the balance; the other nine tenths burn,
+    /// so recycled Sybil reserves keep a real marginal cost.
+    function _bondRecovery()
+        private
+        view
+        returns (BondDisposition, Tree.Node, address, uint256)
```
guidanoli: I would name the return variables for clarity when reading the signature.

prt/contracts/src/tournament/Tournament.sol
GCdePaula: Done!

prt/contracts/src/tournament/Tournament.sol
GCdePaula: Done!

stephenctw (approved):
I went through the Solidity surface against the review guide (`ITournament` / `Tournament` / `MatchClocks`, plus the refund docs).

What I checked and am happy with:
- `eliminableAt` lines up with the inclusive `ELIMINATE_BOTH` boundary for both clock shapes, and the emit/cancel path (create, advance, leaf seal, delete, inner seal) looks right
- `bondRecovery` / `tryRecoveringBond` stay on the same arms, including the bounty formula and the failed-recipient retry case
- `tournamentStanding` and `innerResult` share the same finish/expiry authority, so parent reads shouldn’t drift

I also re-read on the latest tip — the reserve-doc payment wording looks consistent with the code, and the earlier review threads on counter placement / named returns look settled.

cartesi-rollups/contracts/foundry.toml
```diff
@@ -14,6 +14,7 @@ remappings = [
     "cartesi-rollups-contracts-3.0.0/=dependencies/cartesi-rollups-contracts-3.0.0-alpha.6/",
     "forge-std-1.9.6/=dependencies/forge-std-1.9.6/",
     "prt-contracts/=../../prt/contracts/src/",
+    "prt-contracts-test/=../../prt/contracts/test/",
```
guidanoli, line 17: I find it strange to have a dedicated mapping for test files.
These mappings will be used for compiling production code.
I would remove this and use `prt-contracts/../test` instead.

Or make `prt-contracts/` map to `../prt/contracts/` and edit imports:
- `prt-contracts/` -> `prt-contracts/src/`
- `prt-contracts-test/` -> `prt-contracts/test/`.

cartesi-rollups/contracts/src/DaveConsensus.sol
```diff
@@ -417,6 +416,24 @@ contract DaveConsensus is IDaveConsensus, ERC165, ApplicationChecker {
...
+        view
+        returns (bool finished, Tree.Node winnerCommitment, Machine.Hash finalMachineStateHash)
+    {
+        ITournament.TournamentStandingView memory standing = tournament.tournamentStanding();
+        if (standing.standing == ITournament.TournamentStanding.ROOT_WINNER) {
+            return (true, standing.candidate, standing.finalState);
+        } else if (standing.standing == ITournament.TournamentStanding.ROOT_FAILED) {
+            revert ITournament.TournamentFailedNoWinner();
+        } else {
+            return (false, Tree.ZERO_NODE, Machine.ZERO_STATE);
```
guidanoli, line 433: You can suppress these lines if you want.

Given that all return values are named, they are already initialized with their respective default values: `false`, `Tree.ZERO_NODE`, and `Machine.ZERO_STATE`.

GCdePaula: Got it! Stylistically, I prefer the complete if-then-else with an explicit return. I'll leave as is.

cartesi-rollups/contracts/foundry.toml
GCdePaula: This remapping is referenced only by the two Rollups test fixtures. It does not add PRT test sources to the production dependency graph (remappings only resolve imports that are actually used). I prefer the explicit test-support namespace over escaping the source mapping through `prt-contracts/../test`. I'm also not a fan of "re-rooting" the package. I would keep this as-is.

justfile
```diff
@@ -113,12 +128,50 @@ doctor:
...
     if command -v docker > /dev/null; then ok "docker"; else
       warn "docker not on PATH" "needed only for the honeypot image and dockerized workflows"; fi
+    if command -v xgenext2fs > /dev/null; then ok "xgenext2fs"; else
+      warn "xgenext2fs not on PATH" \
+        "needed only to build the honeypot image (its project generates rootfs from a tarball with it)"; fi
+    # Forge formatter heuristics drift across releases; a local/CI
+    # version split fails CI fmt with no local reproduction. Compare
+    # against the setup-tools pin instead of documenting the advice.
+    if command -v forge > /dev/null; then
+      ci_pin=$(sed -n "s/.*default: 'v\([0-9][0-9.]*\)'.*/\1/p" .github/actions/setup-tools/action.yml | head -1)
```
guidanoli: _Suggestion:_ Define the Foundry version in the root `justfile`, and have the CI retrieve it through a `just` recipe (e.g. `print-foundry-version`).

That's what we did on rollups-contracts:
- `Makefile` ([`FOUNDRY_VERSION`](https://github.com/cartesi/rollups-contracts/blob/v3.0.0-alpha.8/Makefile#L58), [`print-foundry-version`](https://github.com/cartesi/rollups-contracts/blob/v3.0.0-alpha.8/Makefile#L389-L390), [`check-foundry-version`](https://github.com/cartesi/rollups-contracts/blob/v3.0.0-alpha.8/Makefile#L365-L384))
- `.github/actions/setup/action.yml` ([`set-foundry-version`](https://github.com/cartesi/rollups-contracts/blob/v3.0.0-alpha.8/.github/actions/setup/action.yml#L7-L10))

prt/contracts/src/ITournament.sol
```diff
@@ -42,14 +46,12 @@ interface ITournament {
...
     /// @dev A root tournament is at level 0.
-    /// A single-level tournament has 1 level.
-    /// A multi-level tournament has 2 or more levels.
     /// Time is measured by the contract time source, currently `block.number`.
     /// For root tournaments (level == 0), nestedDispute fields are zero.
     struct TournamentArguments {
         Commitment.Arguments commitmentArgs;
         uint64 level;
-        uint64 levels;
+        TournamentKind kind;
```
guidanoli, line 54: _Suggestion:_ Add `levels` to `TournamentArguments` and `TournamentDescriptor`, as it tells how close the dispute is to the leaf level, which cannot be inferred from `level` and `kind`.

justfile
GCdePaula: Good suggestion. I moved the expected Foundry version to `FOUNDRY_VERSION` in the root `justfile`, added `just print-foundry-version`, and changed `setup-tools` to install Just first and obtain the version through that recipe. `doctor` now consumes the same root value, so the redundant action input and YAML parsing are gone. Is this what you had in mind?

prt/contracts/src/ITournament.sol
GCdePaula: You’re right that `(level, kind)` alone does not expose the dispute-wide depth. We still prefer not to add `levels` back to every `TournamentArguments` and descriptor: it is factory-wide configuration, would be repeated in every clone, and would give clients a second encoding of leafness to reconcile with `kind`.

Instead, we made the configuration introspectable at its natural owner. `IMultiLevelTournamentFactory` now exposes `tournamentLevelCount()` and `tournamentParameters(level)`. The returned `TournamentParameters` intentionally retains its `levels` field, and root/inner instantiation consumes that same provider-backed view.

This lets consumers fetch the complete prospective geometry from the factory without enlarging every clone. The node now reads row zero during startup and checks its stride and span against the compiled machine geometry.

justfile
```diff
@@ -21,10 +21,16 @@ set shell := ["bash", "-o", "pipefail", "-cu"]
 # rejoining them with spaces (which destroys shell quoting).
 set positional-arguments
 
+FOUNDRY_VERSION := "v1.5.1"
```
guidanoli: I think you can safely drop the `v`. The foundry-rs/foundry-toolchain action (since [1.7.0](https://github.com/foundry-rs/foundry-toolchain/releases/tag/v1.7.0)) accepts version strings without the `v` prefix, as well as foundryup. Plus, this would make dave and rollups-contracts have similarly-named devnet artifact names (both ending in `-anvil-1.5.1.tar.gz`).

justfile
```diff
@@ -159,16 +165,17 @@ doctor:
         "needed only to build the honeypot image (its project generates rootfs from a tarball with it)"; fi
     # Forge formatter heuristics drift across releases; a local/CI
     # version split fails CI fmt with no local reproduction. Compare
-    # against the setup-tools pin instead of documenting the advice.
+    # against the root pin that setup-tools also consumes.
     if command -v forge > /dev/null; then
-      ci_pin=$(sed -n "s/.*default: 'v\([0-9][0-9.]*\)'.*/\1/p" .github/actions/setup-tools/action.yml | head -1)
+      foundry_pin="{{FOUNDRY_VERSION}}"
+      expected_forge="${foundry_pin#v}"
```
guidanoli: If we remove the `v` prefix from the Foundry pin, this can be simplified.

guidanoli (commented):
Looking great! Just raising one minor nit.

prt/contracts/src/ITournament.sol
guidanoli: That is a great way to solve this issue as well!

justfile
guidanoli: Yes, that is exactly what I had in mind. Thank you.

justfile
GCdePaula: Done!

justfile
GCdePaula: Done!

guidanoli (approved):
LGTM 🚀

PR #275, opened 2026-08-08
Security: PoC for permanent tournament freeze (winMatchByTimeout clock-deduction bug)

CharaD7:
## Summary

This PR adds a Foundry PoC test and a disclosure report for a **liveness bug** in the PRT tournament's `winMatchByTimeout` function that can permanently freeze a tournament (and by extension, the application it secures).

## The Bug

In `Tournament.sol::winMatchByTimeout` (lines 198 and 212), when a match is won on timeout, the contract deducts the opponent's elapsed overtime from the winner's clock:

```solidity
_clockOne.deducted(_clockTwo.timeSinceTimeout());
```

`Clock::timeSinceTimeout` grows **without bound** after the opponent's clock expires. `Clock::deducted` uses saturating subtraction (`monus`). When the winner waits past its own remaining allowance after the opponent timed out, the deduction saturates the winner's clock to zero, and `_setNewPaused` reverts with `"can't create clock with zero time"`.

Because the overtime only grows, this revert is **permanent** — `winMatchByTimeout` can never succeed again for this match.

The only remaining resolution path, `eliminateMatchByTimeout`, then deletes the match with `WinnerCommitment.NONE` — both bonds are destroyed and there is no winner. Once the tournament closes, `arbitrationResult` reverts `TournamentFailedNoWinner`, making `DaveConsensus.settle()` impossible forever. The application and its escrowed funds are frozen.

## Proof of Contract

`FreezeBugTest.sol` reproduces the full chain:
1. Two commitments join a tournament; one goes silent.
2. After enough blocks elapse, `winMatchByTimeout` reverts with `"can't create clock with zero time"`.
3. Waiting further blocks — the revert is permanent (overtime only grows).
4. `eliminateMatchByTimeout` deletes the match with no winner.
5. `arbitrationResult` then reverts `TournamentFailedNoWinner`.

```
Running 2 tests for test/FreezeBug.t.sol:FreezeBugTest
[PASS] testClockDeductionFreeze()
[PASS] testClockDeductionFreezeIsPermanent()
```

## Scope & Impact

- **Type:** Liveness / Denial-of-Service (permanent freeze).
- **Severity:** Medium — this is a freeze, NOT a fund-theft vector. No funds can be extracted.
- The winner must wait past its own remaining allowance after the opponent times out for the freeze to trigger; the defect is that this state is irrecoverable.
- A fix should bound the deduction (clamp to a minimum of 1) or ensure a recoverable arbitration path always exists.

## Disclosure

Full analysis, root cause with code references, and recommended fix are in `DISCLOSURE_REPORT.md`.

---

*No on-chain transaction was submitted. This bug was identified through static audit of the public source and verified against deployed bytecode. The withdrawal destination in the honeypot app is a compile-time constant; this bug cannot be used to redirect funds.*

### DIFF
--- a/DISCLOSURE_REPORT.md
+++ b/DISCLOSURE_REPORT.md
@@ -1,0 +1,182 @@
+# Security Disclosure Report
+
+## PRT Tournament Liveness Bug — Permanent App Freeze via `winMatchByTimeout` Clock Deduction
+
+| Field | Value |
+|---|---|
+| **Date** | 2026-08-08 |
+| **Target** | Cartesi PRT Honeypot v3 (Ethereum Mainnet) |
+| **Component** | `dave` — PRT tournament contracts (`cartesi/dave` @ v2.0.0) |
+| **Affected contract** | `Tournament.sol` (deployed via `MultiLevelTournamentFactory`) |
+| **Severity** | **Medium — Liveness / Denial-of-Service** (permanent freeze, NOT fund theft) |
+| **Classification** | Liveness failure; no safety/asset-loss impact |
+| **Researcher** | Independent security research (AI-assisted) |
+| **Status** | Full audit conducted; no extraction path exists. Report submitted in good faith. |
+
+---
+
+## 1. Executive Summary
+
+The Permissionless Refereed Tournament (PRT) dispute system contains a **liveness bug** in `winMatchByTimeout`. When a match winner collects a timeout victory, the contract deducts the opponent's *unbounded* overtime from the winner's own clock. If the winner is slow to claim the win — specifically, slower than their own remaining allowance after the opponent timed out — the deduction saturates the winner's clock to zero and the call **reverts permanently**.
+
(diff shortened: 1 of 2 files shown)


GCdePaula:
Thanks for the report and reproduction.

This identifies a real inconsistency in the legacy implementation on `main`: `canWinMatchByTimeout()` can report that a match is winnable even though `winMatchByTimeout()` then reverts when the overdue charge would exhaust the survivor’s allowance.

We independently found and fixed this in [#273](https://github.com/cartesi/dave/pull/273), now on `next/3.0`. The fix introduces a shared timeout classifier used by the capability view and both timeout mutations:

- The survivor wins only when `remaining > overdue`.
- At equality or below, `eliminateMatchByTimeout()` eliminates both commitments.

Consequently, the legacy match is not permanently stuck: elimination remains available. A root tournament may then finish without a winner, but that is the intentional fail-closed outcome once every remaining commitment has exhausted its clock budget.

We therefore would not adopt the suggested clamp-to-one behavior, since that would grant new time to an expired commitment and violate clock conservation.

The exact boundary is already covered by focused integration tests and fuzz tests in #273, so we will close this PR as already addressed rather than merge the additional PoC into `main`.

Thank you again for taking the time to investigate and report it.

PR #278, opened 2026-08-21
ci: npx actions-up

endersonmaia:


guidanoli (approved):
Commit SHAs and tags match, and bumps are safe.

PR #281, opened 2026-08-21
Bump rollups-contracts to 3.0.0-alpha.9

guidanoli:
Bumps `rollups-contracts` from 3.0.0-alpha.6 to 3.0.0-alpha.9 and adapts the project to the upstream changes, plus a couple of small unrelated fixes bundled in.

**Key changes:**
- Adds `DaveConsensus.wasInputFinalized`, required by the updated `IOutputsMerkleRootValidator` interface (used for deposit refunds). Compares `inputIndex` against the current sealed epoch's input index lower bound; ignores the block-number argument. Covered by new unit tests.
- `DaveAppFactory` now passes `IInputBox` directly to `IApplicationFactory` instead of encoding a DA blob — internal simplification, no API or behavior change for users of `DaveAppFactory` or `DaveConsensus`.
- `BaseDeploymentScript` now also stores deployments in TXT format (JSON deployment artifacts are kept temporarily to help clients migrate but are deprecated); internal consumers switched to reading TXT, letting `blockchain-reader` drop its `serde` dependency.
- Renames `ERC*` → `Erc*` to match upstream renaming (e.g. `ERC20Portal` → `Erc20Portal`).
- Tests now use rollups-contracts' deployment fixture directly (required an upstream import-path fix and local `always_use_create_2_factory` configuration).
- Formatting cleanup: single-line imports across contracts, enabled by a new Foundry 1.5.0 option — no logic changes.
- Docker fix: test image was missing `contract-bindings.sh` copy, breaking `just bind`; gave binding generation its own layer. Unrelated to the version bump, and untested by CI since it does not build this image.

### Review guidance
- Focus on `wasInputFinalized` and its tests.
- Skim the rename/formatting changes — mechanical, low risk.
- Double-check the TXT/JSON dual-write in `BaseDeploymentScript` doesn't break existing JSON consumers during the deprecation window.
- The Docker fix is worth a manual build check since CI doesn't cover it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1736,7 +1736,6 @@ dependencies = [
  "ruint",
  "rusqlite",
  "rusqlite_migration",
- "serde",
  "serde_json",
  "tempfile",
  "testcontainers-modules",
--- a/cartesi-rollups/contracts/foundry.toml
+++ b/cartesi-rollups/contracts/foundry.toml
@@ -32,6 +32,7 @@ recursive_deps = true
 
 [fmt]
 sort_imports = true
+single_line_imports = true
 
 [lint]
 exclude_lints = ["incorrect-shift"]
--- a/cartesi-rollups/contracts/foundry.toml
+++ b/cartesi-rollups/contracts/foundry.toml
@@ -8,21 +8,22 @@ via_ir = true
 
 allow_paths = ["../../prt/contracts", "../../machine/step"]
 remappings = [
-    "@openzeppelin-contracts-5.2.0/=dependencies/cartesi-rollups-contracts-3.0.0-alpha.6/dependencies/@openzeppelin-contracts-5.2.0/",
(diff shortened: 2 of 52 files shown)


GCdePaula (approved):
LGTM!

stephenctw (approved):
Two small optional notes, neither blocking:

Worth a quick manual build of the test Docker image when you get a chance — the bind-layer fix looks correct, but CI doesn’t cover that image.
_writeTxtDeployment says no trailing newline; worth a quick check that writeFile/toString actually guarantee that if anything besides parseAddress reads those files.
Otherwise good to merge from my side.

guidanoli:
> Worth a quick manual build of the test Docker image when you get a chance

Can confirm the test Docker image built successfully locally.

> `_writeTxtDeployment` says no trailing newline;
> worth a quick check that `writeFile`/`toString` actually guarantee that
> if anything besides parseAddress reads those files

Can confirm both functions produce no dangling newline character.

PR #282, opened 2026-08-24
PRT machine-yield check

guidanoli:
This PR alters the signature of `IDaveConsensus.stageTournamentResult` to receive a machine-validity proof (which encompasses not only the start of the tx buffer, where the outputs Merkle root is written, but also the `iflags.Y` and `htif.tohost` registers, which are necessary to check whether the machine is manually yielded with an rx accepted reason). If the tournament result is that of a machine that is not manually yielded or with the wrong reason, then `DaveConsensus` will not allow such a post-epoch machine state to be staged. This prevents ill-formed machine states from ever being finalized on-chain. In such cases, the guardian may foreclose the application, releasing user funds through the emergency withdrawal and deposit refunds mechanisms.

Implementation-wise, the production source contracts leverage the `MachineValidityProof` struct and `LibMachineValidityProof` library from the upstream `rollups-contracts` dependency, and unit tests leverage the upstream `LibEmulator` test utility.

:construction: **Why is this PR in draft?** This PR does not yet contain the necessary adjustments on the Sling node. Among the foreseen changes are: storing `iflags.Y` and `htif.tohost` Merkle proofs alongside the (already-stored) tx buffer proof, and providing such proofs to the `stageTournamentResult` function call.


GCdePaula (approved):
LGTM!

PR #283, opened 2026-08-26
feat!(node): stage machine validity proofs

GCdePaula:
Implements the Sling node side of #282.

The node now captures and persists the Cartesi Machine proofs for `iflags.Y`, `htif.tohost`, and the TX buffer, validates them against the final machine state, and submits them when staging a tournament result.

It also replaces SQLite migration machinery with a create-only schema identified by the node version and raw schema fingerprint. Tests cover real accepted, rejected, and non-yielded machines.


guidanoli (approved):

stephenctw (approved):
LGTM!

PR #284, opened 2026-08-27
Tournament read-interface additions for the reference node

GCdePaula:
Adaptations to the tournament read API requested by the rollups-node team's alpha.4 review, plus the gas recalibration that review work surfaced. The production Solidity footprint is deliberately small. Diffstat over non-test contract sources: **6 files changed, +159/−25**.

## Read surface

- `tournamentStanding()` gains `finishedAt` (accepted earlier), populates `finalState` for `INNER_WINNER`, and gains `winnerExpiresAt` (the first inclusive instant the inner winner becomes eliminable; fixed once the tournament finishes).
- New `commitmentStanding(root)` per-commitment view: join record, claimer, and a clock snapshot with no per-block-varying field. A projection, not the alpha.3 `getCommitment`: raw `Clock.State` stays off the wire and eliminated commitments keep stale snapshots (liveness stays topological).
- `MatchAdvanced` gains `segmentStartPosition`, so a finished match's bisection trajectory survives match deletion in the event history (topic0 changes; only possible pre-release).
- `TournamentDescriptor` gains `startInstant` and `allowance` (the clone's immutable time envelope; joins close at `startInstant + allowance`, inclusive).
- `canStageTournamentResult()` is total: `ROOT_FAILED` answers through a new `isTournamentFailed` flag instead of reverting; `stageTournamentResult` keeps the revert.
- ERC-165 on `Tournament` (through the clones) and `MultiLevelTournamentFactory`.

## Gas recalibration

`Gas.WIN_LEAF_MATCH` 4,420,000 -> 3,885,000: the old value predated the machine-yield check and the alpha-9 bump, which made the maximum-input leaf-proof path cheaper; the exact-relationship witnesses had been failing. Measured under the pinned release environment; dated record in `docs/reviews/2026-08-27-prt-leaf-gas-recalibration/`. Leaf bonds shrink accordingly (about 0.027 ETH at the height examples). The reproducible leaf gate's dependency-digest pin, stale since alpha 9, is fixed alongside.

### DIFF
--- a/cartesi-rollups/contracts/script/measure-prt-leaf-gas.sh
+++ b/cartesi-rollups/contracts/script/measure-prt-leaf-gas.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 readonly EXPECTED_FOUNDRY_VERSION="1.5.1-v1.5.1"
-readonly EXPECTED_DEPENDENCIES_SHA256="bf5c94f033883d49e851fe57111f5031bfbbc1969c6027aedc6ac607815d4234"
+readonly EXPECTED_DEPENDENCIES_SHA256="0390394d7559329a94913a96b298a798c16fb03446600ca746760d5942ae6f4d"
 readonly EXPECTED_MACHINE_HASH="9b358eac8ebd2aa2c7ab4c00d098da7fd90906dc571ec83ec16e889fd220e0fb"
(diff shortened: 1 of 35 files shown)


guidanoli (dismissed):
Solid changes! In particular, I really like that `tournamentStanding` no longer reverts for no-winner tournaments, and instead returns an extra Boolean value signaling failure.

stephenctw (approved):
LGTM!

guidanoli (approved):
LGTM!

PR #285, opened 2026-08-31
chore: bump rollups-contracts to 3.0.0-alpha.10

guidanoli:
This PR bumps the `cartesi-rollups-contracts` Soldeer dependency from `3.0.0-alpha.9` to `3.0.0-alpha.10` (which doesn't actually lead to any smart contract changes, just compilation artifact distribution), and makes the necessary adjustments in the repo related to that bump.

It also bumps the `cartesi-rollups-contracts` Rust dependency from `3.0.0-alpha.4` (which used alloy 1) to `3.0.0-alpha.10` (which, since `3.0.0-alpha.8` uses alloy 2). With this bump, some changes to the Rust node were necessary, such as to switch from `Application` to `IApplication`  and to make small adjustments related to the alloy 1-to-2 migration.

Pertaining to this last bump, it was also necessary to bump `rustc` from 1.90 to 1.94, which is a minimal requirement for alloy 2. We also took the opportunity to remove the `pnpm` version from the test Dockerfile, as it was no longer being used, and to use the interface `IInputBox` rather than `InputBox`. We also now generate Rust bindings for PRT and PRT Rollups contracts using alloy 2.

### DIFF
--- a/.github/actions/setup-tools/action.yml
+++ b/.github/actions/setup-tools/action.yml
@@ -8,9 +8,9 @@ inputs:
   rust-version:
     description: 'Rust version to install (if Rust is setup)'
     required: false
-    # Keep in step with the release container (rust:1.90-trixie in
+    # Keep in step with the release container (rust:1.94-trixie in
     # build.yml); the workspace uses post-1.86 std APIs.
-    default: '1.90'
+    default: '1.94'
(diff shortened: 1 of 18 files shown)


GCdePaula (approved):
LGTM!

